PR #2, opened 2021-11-08
Update README.md

richb-hanover:
I think I incorporated all the comments, but I was moving fast. Feel free to edit anything here. Thanks!


richb-hanover:
Uh oh... I left a question in the "Service" section. Please check my commands. Now off to work! Thanks.

PR #5, opened 2021-11-23
Replace bc with awk

dim-geo:
let's avoid bc in openwrt and use awk everywhere


dim-geo:
Yes, I am familiar, that's why I stepped in ;)
I think now, via this call_awk you can easily implement any algorithm

lynxthecat:
Great work. Looks good to me.

PR #10, opened 2021-12-11
Removed unnecessary tail call.

Fail-Safe:
This tail call is unnecessary if using the most recent version of hping_parser.awk.


lynxthecat:
Thanks for this. Sorry that I forgot to remove earlier.

PR #20, opened 2022-04-06
Add delay for VPN usage

TalalMash:
Script exits with "cannot find ../statistics/tx_bytes" errors on boot in conjunction with a VPN.

### DIFF
--- a/config.sh
+++ b/config.sh
@@ -8,6 +8,9 @@
 # Author: @Lynx (OpenWrt forum)
 # Inspiration taken from: @moeller0 (OpenWrt forum)
 
+#Delay for interface up (e.g Wireguard/OpenVPN)
+sleep 15
+
 # *** OUTPUT OPTIONS ***
 
 output_processing_stats=1 # enable (1) or disable (0) output monitoring lines showing processing stats


TalalMash:
Scratch that, commented procd respawn seems to be effective. Apologies.

lynxthecat:
No need to apologise. Your comments are helpful and I predicted this might happen. 

You see I added in checks to verify the interface files are present before attempting to read, and exiting on startup if they are not and during normal run simply maintaining the previously read rx and tx bytes value. This seems like the right approach. But I may be mistaken. 

@TalalMash are you running ordinary SQM scripts or my cake-wg-pbr script here: 

https://github.com/lynxthecat/cake-wg-pbr

@moeller0 any ideas here? Delay to startup for boot to allow VPN to settle seems like a reasonable idea? Should the delay perhaps be conditional on the detection that script is not being run on a terminal? Or just delay for any startup? 

Otherwise @TalalMash your observation about bufferbloat being detected for an hour sounds very broken to me. Could you please post some data to show what you mean? You may have exposed a flaw with the implementation that needs resolving.

TalalMash:
Hello!
Thank you for the info! I was using oridinary script, and it turns out that there is packet collision when using MACVLAN on boot affecting ping readings in the first few seconds.
I'll try the wg-pbr and report back!

lynxthecat:
Please do report back. I'm continually looking for ways to improve.

lynxthecat:
In fact, I wonder if you would mind running a modification I am testing myself? In essence I tried distinguishing between bufferbloat that is related to high load and bufferbloat that is not. I think on balance I don't like the change, but I'd be grateful if you might test and give me your thoughts?

TalalMash:
> In fact, I wonder if you would mind running a modification I am testing myself? In essence I tried distinguishing between bufferbloat that is related to high load and bufferbloat that is not. I think on balance I don't like the change, but I'd be grateful if you might test and give me your thoughts?
> 
> https://pastebin.com/xCC3fhWn

Thanks, this seems to be less conservative in the beginning when it comes to throttling speed and minimizing latency increase. It also has much faster bandwidth ramp-up time after running for a while from idling, I think the original works better than this script at 40ms RTT max connections, while the newer at 90-100ms seems to be much better. My connection (LTE) does around ~30mbit, I sampled around 10 times. The new script no longer show "bufferbloat" on idle with Wireguard.

lynxthecat:
I think the new script is trying to be too clever. An issue is that the load estimate is just an estimate because we don't know the true rate. So actually when it states 40 it could really be 80. So changing rate decisions based on the load estimate is perhaps wrong. 

About the VPN issue do we need to delay the startup of the script via the service file to allow for WireGuard to settle?

TalalMash:
> Thanks a lot for testing. I think the new script is perhaps trying to be too clever. An issue is that the load estimate is just an estimate because we don't know the true achievable capacity. So actually when it states 40 it could really be 80. The thinking I had was that unless the connection is fairly saturated we shouldn't be reducing the rate based on the achieved rate. But maybe to use this approach I should make it based on load being less than a new percentage. The high load threshold is 75. This new percentage could be say 10. But I'm really not sure about this.
> 
> About the VPN issue do we need to delay the startup of the script via the service file to allow for WireGuard to settle?

Yes that would be a good idea, I'm currently using in the script ping until instead of delay:
```
. $install_dir"config.sh"
until ping -c5 -I $ul_if 8.8.8.8 >/dev/null 2>&1; do :; done
```

lynxthecat:
Ah that's a funky idea I think I'll adopt this for both $ul_if and $dl_if. Make sense?

PR #53, opened 2022-10-03
Add a simple log file parser and ploting script.

moeller0:
To make sense of recorded log files it helps to plot these. This tries to add a simple plotting function written in octave/matlab.

### DIFF
--- a/fn_parse_autorate_log.m
+++ b/fn_parse_autorate_log.m
@@ -11,7 +11,7 @@ function [ autorate_log ] = fn_parse_autorate_log( log_FQN )
 	% dissect the fully qualified name
 	[log_dir, log_name, log_ext ] = fileparts(log_FQN);
 
-	align_rate_and_delay_zeros = 0; % so that delay and rate 0s are aligned, not implemented yet
+	align_rate_and_delay_zeros = 1; % so that delay and rate 0s are aligned, not implemented yet
 	output_format_extension = '.pdf';
 	line_width = 1.0;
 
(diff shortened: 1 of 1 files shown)


moeller0:
@Lynx, have a look whether you like it. Since it does not interact with anything but log files it should be relatively safe to add. Nobody is forced to use it, but since @patrakov asked for sharing, I thought inside CAKE-autorate might be the best place.

patrakov:
Please add a few comments at the top how to use. I think there will be some users who have never encountered Octave before :)

lynxthecat:
Thanks a lot @moeller0 I will take a look tonight. I think what would be
super helpful would be some form of analysis metric to somehow measure how
well cake-autorate (and competitors) is doing. Can you think of any such
metrics we can use?

patrakov:
OK, let me brainstorm the metrics:

* total length of the log in seconds
* total length of non-idle UL/DL in seconds
* total length of sleep
* total number of pings lost
* total number of bad events such as stalls and "response over 500 ms old" (which often happens after stalls)
* total bufferbloat duration in seconds
* min/max/average achieved UL/DL rates (over non-idle periods only)
* min/5%/25%/50%/75%/95%/max shaper rate percentiles (both over non-idle periods only and including idle/sleep time)
* min/5%/25%/50%/75%/95%/max latency percentiles
* average UL/DL bandwidth drop in percent after each bufferbloat event

But again, please keep the scope limited. This kind of analysis belongs in a different MR.

fn_parse_autorate_log.m
```diff
@@ -5,6 +5,14 @@
 	%
 	%       Copyright (C) 2022 Sebastian Moeller
 
+	% HOWTO:
+	% you need to install octave (https://octave.org)
+	% then navigate to the directory containing fn_parse_autorate_log.m and either:
+	% run 'octave --gui' in a terminal and open the fle and run it (recommended if you want/need to edit values)
```
patrakov, line 11: Typo: file

fn_parse_autorate_log.m
```diff
@@ -5,6 +5,14 @@
 	%
 	%       Copyright (C) 2022 Sebastian Moeller
 
+	% HOWTO:
+	% you need to install octave (https://octave.org)
+	% then navigate to the directory containing fn_parse_autorate_log.m and either:
+	% run 'octave --gui' in a terminal and open the fle and run it (recommended if you want/need to edit values)
+	% or run 'octave ./fn_parse_autorate_log.m' from the terminal
+	% by default the code will open a file selection dialog which should be used to select a CAKE-autorate log file.
```
patrakov, line 13: Is there a fully non-interactive way to convert a log file to a png or svg with graphs?

moeller0:
I think trying to come up with metrics is a different issue.
I have not thought about how to compare different approaches without being able to recreate the exact identical traffic conditions.

PR #55, opened 2022-10-03
Octave plotter

moeller0:
Fixed version of the plotting helper:
The following invocation will work non-interactively:
`octave -qf --eval 'fn_parse_autorate_log("./SCRATCH/cake-autorate.log.20221001_1724_RRUL_fast.com.log", "./out.png")'`
First argument is the input log file, the second optional argument is the outputfile, the file type is detected from the extension, `.png` in the example, known types are:
```
.pdf
.png
.tif
.ps
```

BTW: Sorry for the lost PR, I had srewed up my local git repository and in the course of fixing that I deleted the old branch and had to re-create it.

### DIFF
--- a/fn_parse_autorate_log.m
+++ b/fn_parse_autorate_log.m
@@ -11,7 +11,7 @@ function [ autorate_log ] = fn_parse_autorate_log( log_FQN )
 	% dissect the fully qualified name
 	[log_dir, log_name, log_ext ] = fileparts(log_FQN);
 
-	align_rate_and_delay_zeros = 0; % so that delay and rate 0s are aligned, not implemented yet
+	align_rate_and_delay_zeros = 1; % so that delay and rate 0s are aligned, not implemented yet
 	output_format_extension = '.pdf';
 	line_width = 1.0;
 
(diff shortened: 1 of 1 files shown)


lynxthecat:
Sweet! Have you tried making a couple of example plots on the data you
collected @moeller0?

moeller0:
So, it seems to work on my three testing system and my two log files.
I had to add detection for the legacy gnuplot backend (as that does not allow fancy legend formatting).
But with that fix in place it now seems to work over a SSH connection and produces a usable plot.
When used directly on a computer it cab default to a more capable backend like `qt` and the formatting is
a bit nicer.

moeller0:
As for an example, I posted one to the CAKE w/ Adaptive Bandwidth thread:
[

](url)

lynxthecat:
Wow @moeller0 did you write this from scratch? This looks like a ton of work.

moeller0:
This is mostly new, I included a few helper functions I had sitting around (no need to re-invent the wheel twice) but the parser happened over yesterday, and a few refinements bug fixes today. As I said, I have written a few shoddy parsers already in the past so this was not a completely foreign challenge.

PR #66, opened 2022-10-30
Switch plotting of achieved rates from DATA to LOAD record information

moeller0:
This will now plot achieved rates during sleep periods which might help to confirm wake from sleep happens under the expected/configured conditions. This also fixes small issues (like actually reading in the LOAD records instead of just silently ignoring them).
This also switches the range definition from sample indices to time in second. Since log file start as DATA and LOAD data have independent time axis, there is no guaranteed common time index. Also time based ranges are more intuitive. The code will select the next best sample for each end of the range so specification does not need to be overloy precise. This range can now be specified when calling from
the command line as optional 3rd argument.


moeller0:
This seems to work now and actually used LOAD data if available and falls back to getting the achieved rates from the DATA records otherwise.

PR #67, opened 2022-10-30
Plot cd fs by load percentage

moeller0:
Try to plot something interesting in regards to high and ow load conditions.


lynxthecat:
Nice work. Will have an opportunity to test this extensively when my Zyxel NR7101 gets installed outside and it is time to optimise my cake-autorate parameters.

moeller0:
So having slept over it, I am sure the definition of high and low load is sub-optimal (e.g. because in observation only mode we have no meaningful load percentages, let alone the problem with our load percentage not being a veridical measure of true capacity use over the bottleneck ;)). I am open for other ideas here.
Also treating RTTs as OWDs is a bit off, since we split the RTT into two OWDs artificially and equally, so ideally we need to add the two OWDs to get back to the RTTs... maybe we should do this unconditionally as additional column.

PR #78, opened 2022-12-07
Plotting changes 03

moeller0:


moeller0:
This is intended to not error out for log files without DATA records... the only surviving subplot shows just achieved UL and DL rates and is IMHO not really useful, but plotting something clearly is better than just erroring out...
(if we could also add the two shaper rate settings into the LOAD records, these plots would be a bit less useless, but without delay data I am not sure how interpretable they actually are for more than debugging potential cases where wake-up of the pingers is broken). I wonder whether we simply could wake up the pingers every 5 minutes or so, collect a sample and put them to sleep again, so there are always some DATA samples? Or just force a wake-up when log export is requested and delay the actual export for enough time to collect at least on delay sample?

lynxthecat:
OK so add shaper rates to LOAD records (which we originally put in so that you had data during sleeps) and timeout the wait for load resumption to 5 mins (user configurable)?

PR #85, opened 2022-12-17
Log errors to systemlog

moeller0:
Success will be loged to the log file, failure to the log file and the system log


lynxthecat:
I think I have added in these aspects in other commits. But let me know if there's anything else we should add in that I missed.

PR #86, opened 2022-12-26
Testing mod 20221220

moeller0:
So this is the version I am currently testing (I made some changes today so this needs a bit more baking).
Note I upped the verbosity of the launcher and switched to one-by-one termination (required for better verbosity).

While I consider your ERROR logging approach both elegant and opaque, I think the cmd_wrapper approach is more appropriate as for debugging I want to see both failure and success reports. Note how SILENCE_CMD can be used to toggle cmd_wrapper to demote expected errors to debug messages instead, still even when expected logging them seems helpful.

This is mainly intended to give you visibility on the "hack and slash" I submitted your code to, sorry ;)


moeller0:
Wrong branch, not intended for main....

PR #104, opened 2023-01-01
Update CHANGELOG.md to have "Unreleased" section

richb-hanover:
This can describe new work that might be available in the current code, but hasn't seen a full release.

Having the **Unreleased** section allows you to write up the features as they're being checked in, so it's not such a scramble when it's time for a real release.

Feel free to flesh out my mumblings...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ bandwidth settings by measuring traffic load and RTT times.
 Read the [README](./README.md) file for more details.
 This is the history of changes.
 
+## Unreleased - Version 1.2
+
+- Many adjustments to the underlying code (what?)
+- Many tweaks to plotting code (what?)
+- What else?
+
 ## 2022-09-28 - Version 1.1
 
 Implemented several new features such as:


CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ bandwidth settings by measuring traffic load and RTT times.
 Read the [README](./README.md) file for more details.
 This is the history of changes.
 
+## Unreleased - Version 1.2
+
+- Many adjustments to the underlying code (what?)
```
lynxthecat, line 11:
```suggestion
- cake-autorate now includes a sophisticated offline log file analysis utility written in Matlab/Octave: 'fn_parse_autorate_log.m' and maintained by @moeller0 (OpenWrt forum). This utility takes in a cake-autorate generated log file (in compressed or uncompressed format), which can be generated on the fly by sending an appropriate signal, and presents beautiful plots that depict latency and bandwidth over time together with many important cake-autorate vitals. This gratly simplifies assessing the efficacy of cake-autorate and associated settings on a given connection. 
- Multiple instances of cake-autorate is now supported. cake-autorate can now be run on multiple interfaces such as in the case of mwan3 failover. The interface is assigned by designating an appropaite interface identifier 'X' in the config file in the form cake-autorate_config.X.sh. A launcher script has been created that creates one cake-autorate instance per cake-autorate_config file placed inside /root/cake-autorate/. Log files are generated for each instance using the form /var/log/cake-autorate.X.log. The interface identifier 'X' cannot be empty. 
- Improved reflector management. With a relatively high frequency (default 1 minute) cake-autorate now compares reflector baselines and deltas and rotates out reflectors with either baselines that are excessively higher than the minimum or deltas that are too close to the trigger threshold. And with a relatively low frequency (default 60 minutes), cake-autorate now randomly rotates out a reflector from the presently active list. This simple algorithm is intended to converge upon a set of good reflectors from the intitial starting set. The initial starting set is now also randomized from the provided list of reflectors. The user is still encouraged to test the initial reflector list to rule out any particularly far away or highly variable reflectors. 
-  Reflector stats may now optionally be printed to help monitor the efficacy of the reflector management and quality of the present reflectors. 
-  LOAD stats may now optionally be printed to monitor achieved rates during sleep periods when pingers are shutdown.  
-  For each new sample, the baseline is now subtracted after having been updated rather than before having been updated. 
-  Pinger prefix and arguments are now facilitated for the chosen pinger binary to help improve compatibility with mwan3.
-  Consideration was afforded to switching over to the use of SMA rather than EWMA for reflector baselines, but SMA was found to offer minimal improvement as compared to EWMA with appropriately chosen alpha values. The present use of EWMA with multiple alphas for increase and decrease enables tracking of either reflector owd minimums (conservative default) or averages (by setting alphas to around e.g. 0.095). 
-  User can now specify own log path, e.g. in case of logging out to cloud mount using rclone or USB stick
```

lynxthecat: See here: 

https://github.com/lynxthecat/cake-autorate/blob/testing/CHANGELOG.md

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ bandwidth settings by measuring traffic load and RTT times.
...
+- Many tweaks to plotting code (what?)
```
lynxthecat, line 12:
```suggestion
```

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ bandwidth settings by measuring traffic load and RTT times.
...
+- What else?
```
lynxthecat, line 13:
```suggestion
```

lynxthecat (approved):

richb-hanover:
Cool beans! Thanks!

lynxthecat:
I put reviews in before merging without approving reviews. So had to revert then push further commit to main. Net result is that your unreleased section is in. Although how would this work with new branches? Say I create a new branch, do I keep putting changes in main unreleased section and not new branch? But then won't pulling in the new branch overwrite the unreleased section? Not sure how that works in GitHub. 

You see I had already created this: https://github.com/lynxthecat/cake-autorate/blob/testing/CHANGELOG.md

richb-hanover:
Here's my take on the workflow...

The CHANGELOG can always have information about features that are "on the launching pad". That is, they have been committed to `main`, but not tagged as a full release. (And therefore they are available for people who check out `main` as opposed to a tagged version, or a random development branch.)

Work in a branch can proceed unimpeded,  and it makes sense to update its CHANGELOG to reflect that work. When the branch gets committed, there might be some conflicts to resolve in the CHANGELOG, but it'll be straightforward text editing to sort it out.

PR #107, opened 2023-01-08
Add startup messages reporting function name and PID to functions tha…

moeller0:
…t contain loops

All potentially longer running functions should log their start in the log to aid in debugging, also report the PID while at it.


lynxthecat:
Good idea.

PR #108, opened 2023-01-08
Add stopp message to cleanup_and_killall

moeller0:
For debugging it is helpful to know when a specific autorate instance was terminated to add a message to the system log.


lynxthecat:
Looks helpful.

PR #112, opened 2023-01-20
Syslog move start meassge after instance_id instantiation

moeller0:
…to improve the log type tag

Placed at the current position before ${instance_id} is initiated the SYSLOG message looked like:

Jan 20 19:59:25 turris cake-autorate.: SYSLOG: 1674244765.530486 Starting cake-autorate with config: /root/cake-autorate/cake-autorate_config.vdsl2.sh

missing the ${instance_id} as suffix after autorate to disambiguate different instances with this commit this message now looks like expected:

Jan 20 20:01:53 turris cake-autorate.vdsl2: SYSLOG: 1674244913.515046 Starting cake-autorate with config: /root/cake-autorate/cake-autorate_config.vdsl2.sh

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -1211,7 +1211,6 @@ else
 	config_path=/root/cake-autorate/cake-autorate_config.primary.sh
 fi
 
-log_msg "SYSLOG" "Starting cake-autorate with config: ${config_path}"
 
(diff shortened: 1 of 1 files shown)


lynxthecat:
Looks good to me.

PR #113, opened 2023-01-22
Reduce the required memory for octave's CDF/PDF plots

moeller0:
This reduces the memory demany by increasing the CDF granularity from 1 µsec to 5 µsec and by clearing the large CDF data structures after use. This should make plotting feasible again without excessive memory demands.


lynxthecat:
Nice one! Will this significantly reduce memory requirement such that users won't need to create a big virtual ram file?

PR #118, opened 2023-02-07
[setup script] remove cake-autorate_launcher.sh if it already exists

rany2:


lynxthecat:
Looks good. Thanks for this.

PR #121, opened 2023-02-07
 Set trap to true while in the process of gracefully shutting down 

### LINKED ISSUE #120: [testing branch] cake-autorate sometimes doesn't restart properly
It doesn't happen all the time but sometimes it just doesn't restart and gets stuck. Only solution is to pkill -9 it and remove the /var/run/cake-autorate/primary directory and then start again.

```
root@gw:~# ps aux|grep cake
root      3944  0.0  0.0      0     0 ?        Z    18:00   0:00 [cake-autorate.s] <defunct>
root      3946 14.6  0.0   3288  2356 ?        S    18:00   1:06 /bin/bash /root/cake-autorate/cake-autorate.sh /root/cake-autorate/cake-autorate_config.primary.sh
root      3947  0.0  0.0      0     0 ?        Z    18:00   0:00 [cake-autorate.s] <defunct>
root      4158  0.0  0.0   2152  1672 ?        S    17:00   0:00 /bin/bash /root/cake-autorate/cake-autorate_launcher.sh
root      4212 15.0  0.0   3176  2748 ?        S    17:00  10:06 /bin/bash /root/cake-autorate/cake-autorate.sh /root/cake-autorate/cake-autorate_config.primary.sh
root      4244  0.0  0.0   2744  1328 ?        S    17:00   0:00 /bin/bash /root/cake-autorate/cake-autorate.sh /root/cake-autorate/cake-autorate_config.primary.sh
root      4274  4.1  0.0   3080  2108 ?        S    17:00   2:45 /bin/bash /root/cake-autorate/cake-autorate.sh /root/cake-autorate/cake-autorate_config.primary.sh
root      6147  3.1  0.0   3112  2148 ?        S    17:00   2:05 /bin/bash /root/cake-autorate/cake-autorate.sh /root/cake-autorate/cake-autorate_config.primary.sh
root      6149  0.8  0.0   3252  2288 ?        S    17:00   0:32 /bin/bash /root/cake-autorate/cake-autorate.sh /root/cake-autorate/cake-autorate_config.primary.sh
root     11052  0.0  0.0   1144   820 pts/0    S+   18:07   0:00 grep cake
```

rany2:
Fixes https://github.com/lynxthecat/cake-autorate/issues/120

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -21,7 +21,7 @@ trap cleanup_and_killall INT TERM EXIT
 
 cleanup_and_killall()
 {	
-	trap - INT TERM EXIT
+	trap true INT TERM EXIT
 	
 	log_msg "DEBUG" "Starting: ${FUNCNAME[0]} with PID: ${BASHPID}"
 	
@@ -49,6 +49,8 @@ cleanup_and_killall()
 	[[ -d /var/run/cake-autorate ]] && compgen -G /var/run/cake-autorate/* > /dev/null || rm -r /var/run/cake-autorate
 
 	log_msg "SYSLOG" "Stopped cake-autorate with PID: ${BASHPID} and config: ${config_path}"
+
+	trap - INT TERM EXIT
 	exit
 }
 


lynxthecat:
Trying to work out why this is needed. Why does that function end up getting called twice?

rany2:
@lynxthecat my guess is the following:

- OpenWRT's procd tries to restart cake-autorate
- procd sends SIGINT to cake-autorate but cake-autorate takes too long and causes procd to reach a timeout
- procd sends SIGTERM and this time that works because `cleanup_and_killall` removed the trap but in the process it is blocked from gracefully shutting down and killing the child processes

rany2:
it's not confirmed, but I could confirm that this issue happens when spamming ctrl-c or trying to send a kill signal multiple times

rany2:
> Why does that function end up getting more than once?

Sorry I misread your question, the issue is that it doesn't get called multiple times. The trap handler is removed too early but we still want it to handle all kill signals. If you want a cleaner solution, we could have the trap function be `true` when entering, in fact I could do that now.

lynxthecat:
@rany can you help me figure out:

https://forum.openwrt.org/t/cake-w-adaptive-bandwidth/135379/2189?u=lynx

rany2:
p.s. could you confirm if this PR fixes the issue described in the forum?

lynxthecat:
I'll test, but can you help me understand: what is causing cleanup_and_killall() to trap more than once. I mean where does the second exit signal come from? The first is the kill from cake-autorate_launcher.sh. But what causes the second?

rany2:
> So 'trap true' blocks subsequent traps but prevents repeated calls of cleanup_and_killall()?

Yes, it ends up calling the `true` bash built-in which does nothing other than return 0.

lynxthecat:
So I haven't been able to reproduce my issue with your fix. I'd really like to understand what's going on here.

rany2:
I never had "not stopped on SIGTERM, sending SIGKILL" in my log. This is strange, maybe it's something to do with CPU speed?

lynxthecat:
I see this in the maintain_pingers() process:

```
root@OpenWrt:~/cake-autorate# strace -y -p 1026
strace: Process 1026 attached
read(0</tmp/run/cake-autorate/primary/sleep_fifo>, ^Cstrace: Process 1026 detached
```

rany2:
I mean this is a stupid suggestion, but why not send a newline to sleep_fifo as part of kill routine?

rany2:
> That means it's stuck in the read inside sleep_s right? Why would that block on the kill of maintain_pingers inside the cleanup_and_killall()?

Yes

lynxthecat:
Good suggestion, but why is it needed? The sleep_fifo is held open with:

```
mkfifo ${run_path}/sleep_fifo
exec {fd}<> ${run_path}/sleep_fifo
```

rany2:
Sending a newline will forcibly end the sleep and cause it to get unblocked

lynxthecat:
OK but what I don't understand is why does it block in the first place given that it is held open? sleep_s is used all the time in cake-autorate, so why does it only block on cleanup_and_killall()?

PR #123, opened 2023-02-08
Another attempt at fixing service shutdown issue

rany2:
I wasn't able to trigger the issue but I'm struggling to do so even with the current testing branch (even though I should be able to); so I'm kind of trying to fix the issue even though I don't experience. Either way hopefully it works!

I reverted https://github.com/lynxthecat/cake-autorate/commit/8a0e9d3b055a8c2271d5d048ed9e6731c992ede3 because it's too hacky for my liking. I prefer the FIFO approach aesthetics wise. 

Might fix https://github.com/lynxthecat/cake-autorate/issues/120

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -1126,25 +1126,21 @@ verify_ifs_up()
 sleep_s()
 {
 	# calling external sleep binary is slow
-	# bash does have a loadable sleep 
+	# bash does have a loadable sleep
 	# but read's timeout can more portably be exploited and this is apparently even faster anyway
 
 	local sleep_duration_s=${1} # (seconds, e.g. 0.5, 1 or 1.5)
 
-	read -t ${sleep_duration_s} < ${run_path}/sleep_fifo
+	read -t ${sleep_duration_s} <><(:) || :
 }
 
 sleep_us()
 {
-	# calling external sleep binary is slow
-	# bash does have a loadable sleep 
-	# but read's timeout can more portably be exploited and this is apparently even faster anyway
-
 	local sleep_duration_us=${1} # (microseconds)
-	
+
 	sleep_duration_s=000000${sleep_duration_us}
 	sleep_duration_s=$((10#${sleep_duration_s::-6})).${sleep_duration_s: -6}
-	read -t ${sleep_duration_s} < ${run_path}/sleep_fifo
+	sleep_s ${sleep_duration_s}
 }
 
 sleep_remaining_tick_time()
@@ -1314,9 +1310,6 @@ if [[ -d ${run_path} ]]; then
 else
 	mkdir -p ${run_path}
 fi
-
-mkfifo ${run_path}/sleep_fifo
-exec {fd}<> ${run_path}/sleep_fifo
 
 no_reflectors=${#reflectors[@]} 
 
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -17,8 +17,6 @@
 # Possible performance improvement
 export LC_ALL=C
 
-exec {sleep_fd}<> <(true)
-
 trap cleanup_and_killall INT TERM EXIT
 
 cleanup_and_killall()
@@ -49,9 +47,6 @@ cleanup_and_killall()
 
 	[[ -d ${run_path} ]] && rm -r ${run_path}
 	[[ -d /var/run/cake-autorate ]] && compgen -G /var/run/cake-autorate/* > /dev/null || rm -r /var/run/cake-autorate
-
-	# close sleep file descriptor
-	exec {sleep_fd}>&-
 
 	log_msg "SYSLOG" "Stopped cake-autorate with PID: ${BASHPID} and config: ${config_path}"
 
@@ -1147,7 +1142,7 @@ sleep_s()
 
 	local sleep_duration_s=${1} # (seconds, e.g. 0.5, 1 or 1.5)
 
-	read -t ${sleep_duration_s} -u ${sleep_fd}
+	read -t ${sleep_duration_s} < ${run_path}/sleep_fifo
 }
 
 sleep_us()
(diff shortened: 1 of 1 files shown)


cake-autorate.sh
```diff
@@ -30,23 +30,23 @@ cleanup_and_killall()
...
 	fi
 
-	if ! [[ -z ${maintain_log_file_pid} ]]; then
+	if [[ -n ${maintain_log_file_pid} ]]; then
 		log_msg "DEBUG" "Terminating maintain_log_file_pid: ${maintain_log_file_pid}."
 		kill_and_wait_by_pid_name maintain_log_file_pid 0
 	fi
 
 	[[ -d ${run_path} ]] && rm -r ${run_path}
-	[[ -d /var/run/cake-autorate ]] && compgen -G /var/run/cake-autorate/* > /dev/null || rm -r /var/run/cake-autorate
```
lynxthecat, line 51: Ah. But I think this extra test is needed because cake-autorate supports multiple instances running at the same time. With this change stopping one instance would break the other instances?

cake-autorate.sh
```diff
@@ -824,7 +825,7 @@ maintain_pingers()
 	# this initiates the pingers and monitors reflector health, rotating reflectors as necessary
 
  	trap '' INT
-	trap 'terminate_reflector_maintenance=1' TERM EXIT
+	trap 'kill_maintain_pingers' TERM EXIT
```
lynxthecat, line 828: The reason for indirect termination is because replace_reflector kills and restarts fping / ping. So if this strikes right before the pid is updated then we can be left with residual unkilled fping / ping processes.

cake-autorate.sh
rany2: It's worth noting that the `[[ -d .... ]]` at the start doesn't do anything. You could see this in: `false && true || echo Hello.`

rany2: What's the use of compgen in this case?

rany2:
```
$ false && false || echo Hello.
Hello.
$ true && false || echo Hello.
Hello.
$ false && true || echo Hello.
Hello.
```

cake-autorate.sh
rany2: Yes, makes sense. I'll see what to do then..

cake-autorate.sh
lynxthecat: Ah yes I see. I should have square bracketed?. The use of compgen is to verify that no other instances exist, instances are set as subdirectories within /var/run/cake-autorate like /primary/ and /secondary/ as set by the name of the identifier in the config e.g. cake-autorate_config.primary.sh.

rany2: Ah this makes sense, you could use rmdir instead. `rmdir /var/lib/cake-autorate 2>/dev/null`.

rany2: `rmdir` will remove the folder only if it is empty.

rany2:
I saw a similar issue to our blocking online, I'm trying to open the fd in the subprocess now; I think there's a 50% it will work now.

rany2:
Sorry if this PR is starting to have too many moving parts but I wanted to add https://github.com/lynxthecat/cake-autorate/pull/123/commits/62d536640403a41ce8789686615314d01303ef21 because when cake-autorate restarts the log gets rotated and I'm forced to restart tail -f. I believe this approach would fix that, but I'm not entirely sure (didn't test yet).

lynxthecat:
Great looking at the changes now. So grateful to have your input on this now.

cake-autorate.sh
```diff
@@ -32,26 +30,23 @@ cleanup_and_killall()
 	log_msg "INFO" ""
 	log_msg "INFO" "Killing all background processes and cleaning up temporary files."
 
-	if ! [[ -z ${maintain_pingers_pid} ]]; then
+	if [[ -n ${maintain_pingers_pid} ]]; then
```
lynxthecat, line 33: BTW what's the reason to switch over to -n rather than ! -z?

cake-autorate.sh
```diff
@@ -32,26 +30,23 @@ cleanup_and_killall()
...
 		log_msg "DEBUG" "Terminating maintain_log_file_pid: ${maintain_log_file_pid}."
 		kill_and_wait_by_pid_name maintain_log_file_pid 0
 	fi
 
 	[[ -d ${run_path} ]] && rm -r ${run_path}
-	[[ -d /var/run/cake-autorate ]] && compgen -G /var/run/cake-autorate/* > /dev/null || rm -r /var/run/cake-autorate
-
-	# close sleep file descriptor
-	exec {sleep_fd}>&-
+	rmdir /var/run/cake-autorate 2>/dev/null
```
lynxthecat, line 49: I like the idea of using 'rmdir' in this way.

cake-autorate.sh
```diff
@@ -1142,25 +1134,21 @@ verify_ifs_up()
...
 {
 	# calling external sleep binary is slow
-	# bash does have a loadable sleep 
+	# bash does have a loadable sleep
 	# but read's timeout can more portably be exploited and this is apparently even faster anyway
 
 	local sleep_duration_s=${1} # (seconds, e.g. 0.5, 1 or 1.5)
 
-	read -t ${sleep_duration_s} -u ${sleep_fd}
+	read -t ${sleep_duration_s} <><(:) || :
```
lynxthecat, line 1142: How does this compare efficiency wise? Do we need the '|| :'? It seems that:

'read -t 1.0 <><(:)' 

also works?

cake-autorate.sh
rany2: Really there is no reason, it just looks less confusing style-wise

cake-autorate.sh
rany2: yes it does, but `|| :` ensures that the sleep functions won't exit with an error. We could get rid of it but the advantage with this approach is that it will make it easier for us to set correctness options like `set -e` in the future.

cake-autorate.sh
rany2: So no changes needed?

lynxthecat: Yes much more elegant than my crazy compgen idea.

PR #125, opened 2023-02-08
Replace FIFOs with file descriptors and other improvements and fixes

rany2:
* `${reflectors[@]:0:${no_pingers}}` without quotes has the risk of
  being expanded improperly if an array element contains spaces
  (never going to happen but it doesn't hurt to put it in quotes)

* dl_delays and ul_delays are initialized improperly, at startup
  they only have one element because `for i in {1..${bufferbloat_detection_window}};`
  doesn't get expanded into `for i in {1..<some num here>}`..
  (limitation of brace expressions)


rany2:
Don't merge this yet, I'm going to include some more major changes here.

rany2:
I'll stop for now. Sorry I got carried away with this PR but at least now there are no more FIFOs !

rany2:
Check commit msgs for more details, the PR title and description are inaccurate now.

PR #138, opened 2023-02-09
Replace USR2 call with coproc for pause_maintain_pingers

rany2:


rany2:
Doesn't work when backgrounded

PR #158, opened 2023-02-16
Add safety check to avoid bug reports caused by updating while running

rany2:
* Seeing that even developers make this error, it seems wise to protect users from themselves.

### DIFF
--- a/cake-autorate_setup.sh
+++ b/cake-autorate_setup.sh
@@ -12,6 +12,14 @@ SRC_DIR="https://github.com/lynxthecat/cake-autorate/archive/refs/heads/"
 DOC_URL="https://github.com/lynxthecat/CAKE-autorate#installation-on-openwrt"
 BRANCH="master"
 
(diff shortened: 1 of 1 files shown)


cake-autorate_setup.sh
```diff
@@ -12,6 +12,14 @@ SRC_DIR="https://github.com/lynxthecat/cake-autorate/archive/refs/heads/"
 DOC_URL="https://github.com/lynxthecat/CAKE-autorate#installation-on-openwrt"
 BRANCH="master"
 
+# Check if an instance of cake-autorate is already running and exit if so
+if [ -d /var/run/cake-autorate ]; then
+	echo "cake-autorate is already running - exiting" >&2
```
lynxthecat, line 17:
```suggestion
	printf "cake-autorate is already running - exiting\n" >&2
```

lynxthecat: Can we swap these with printf's?

PR #169, opened 2023-03-03
Minor change to export script

rany2:


rany2:
@lynxthecat Untested because I don't have any gateway running OpenWRT at the moment.

PR #172, opened 2023-03-08
Intercept stderr with execs rather than coprocs

rany2:
Edit: UNTESTED because I'm on FreeBSD now but it should work.


rany2:
@lynxthecat Does this work?

lynxthecat:
I imagine it will. Hoping to get an opportunity to test tonight. Wouldn't
it make sense to ditch the while loop and just use single read?

rany2:
Definitely not. An error could be raised in the cleanup function and/or in any subsequent commands that managed to get executed before the main script got the signal to exit. To illustrate my point see the below example:

```bash
#!/bin/bash

cleanup() {
	echo >&2 Cleanup made an oopsie
	echo >&2 Cleanup made an oopsie
	echo >&2 Cleanup made an oopsie
	echo >&2 Cleanup made an oopsie
	echo >&2 Cleanup made an oopsie
}

trap cleanup INT TERM EXIT

exec 2> >(
	while read -r l
	do
		echo ERROR: $l >&2
		kill -INT $$
	done
)

THIS_COMMAND_DOES_NOT_EXIST
echo >&2 This is something else in stderr.

exit 0
```

lynxthecat:
So back when I tried capturing multiple errors like this I saw a hang on
exit relating to the read of the proc file by the proc manager. So I made
sure it only called kill -INT $$ once and breaks out the while loop. I
think trying to capture subsequent errors might be risky. At least in the
version I tried back when I had while loop.

lynxthecat:
Also if we break after first error we are not capturing subsequent errors anyway right? Or how does that work? I've not had chance to test yet.

rany2:
> Also if we break after first error we are not capturing subsequent errors anyway right? Or how does that work? I've not had chance to test yet.

If you break after the first error, the script will exit uncleanly because stderr got closed which is something unacceptable to bash. It is extremely important for that fd to be open throughout the script's operation (even during cleanup)

lynxthecat:
If you have time would you mind placing these functions back in cake-autorate.sh to where the debug function originally was? No worries if you don't have time I'll just get round to it when I get a chance.

rany2:
@lynxthecat I take it that this works fine?

lynxthecat:
I haven't been able to test. My wife has my router password!

rany2:
@lynxthecat HAHAHA! She is finally fed up with you fucking up the network... The power dynamics have changed.

lynxthecat:
Hah. Well I'm head the household, but she's the neck.

lynxthecat:
Indeed something along these lines! I say I'm the head of the household. She says that she's the neck.

lynxthecat:
I tested and it doesn't always work:

```
ERROR; 2023-03-09-21:57:08; 1678399028.600953; ./cake-autorate.sh: line 1358: ERROR_NOW: command not found
ERROR; 2023-03-09-21:57:08; 1678399028.605308; Exiting cake-autorate now.
SYSLOG; 2023-03-09-21:57:08; 1678399028.602876; Starting cake-autorate with PID: 32564 and config: /root/cake-autorate/cake-autorate_config.primary.sh
DEBUG; 2023-03-09-21:57:08; 1678399028.610494; Starting: cleanup_and_killall with PID: 32564
INFO; 2023-03-09-21:57:08; 1678399028.611953; Stopping cake-autorate with PID: 32564 and config: /root/cake-autorate/cake-autorate_config.primary.sh
INFO; 2023-03-09-21:57:08; 1678399028.613281; Killing all background processes and cleaning up temporary files.
ERROR; 2023-03-09-21:57:08; 1678399028.615462; /root/cake-autorate/cake-autorate_lib.sh: line 193: /var/run/cake-autorate/primary/proc_state: No such file or directory
ERROR; 2023-03-09-21:57:08; 1678399028.620556; Exiting cake-autorate now.




^Croot@OpenWrt-1:~/cake-autorate#
```

PR #200, opened 2023-07-01
Remove unused locking functions from lib.sh

rany2:


lynxthecat:
Might not they prove helpful one day? Or perhaps my new fangled IPC mechanism truly renders them obselete?

rany2:
We could always reintroduce it

PR #203, opened 2023-07-02
Drop reference to proc_man from README.md

rany2:
With proc_man gone, this paragraph no longer applies.


lynxthecat:
See here: 

https://github.com/lynxthecat/cake-autorate/blob/e8c3fa4b8cb768d3cf7a096860c7a9125cda65a3/cake-autorate.sh#L2025

rany2:
@lynxthecat Shouldn't a lock be acquired before such an operation?

lynxthecat:
Obviated by design.

PR #206, opened 2023-07-03
Minor changes for the release

rany2:


rany2:
@lynxthecat Could you test this in this state? I won't be pushing any changes anymore

rany2:
Why https://github.com/lynxthecat/cake-autorate/pull/206/commits/2633d365f81b9a7d83871c335459f920728a6d13? It hurts readability and brings no performance improvements

lynxthecat:
I started looking into it from perspective of shaper_rate_kbps_str - the _str jarred. So I thought about making it something like new_shaper_rate_kbps. And then it dawned on me that we have set_shaper_rates() and set_cake_rate(), which also jarred. So I thought actually why not just fold it all into the one function. It's actually simpler and avoids messing about with the assignments and variable names and awkwardness over set_shaper_rates() vs set_cake_rate(). It seems simpler to me to have all of this in just the set_shaper_rates() function?

lynxthecat:
Could be squashed with: https://github.com/rany2/cake-autorate/commit/137d5b05d26df9739e432c3f1fd9a2da7a536097?

rany2:
@lynxthecat Personally I'm not a fan of this but I don't feel strongly about it. I just don't like how you repeated code instead of reusing the same function.

rany2:
> Could be squashed with: [rany2@137d5b0](https://github.com/rany2/cake-autorate/commit/137d5b05d26df9739e432c3f1fd9a2da7a536097)?

Sure

lynxthecat:
Actually you're causing me to doubt now. 

Maybe the old formulation was indeed better, but with `new_shaper_rate_kbps` rather than `shaper_rate_kbps_str`. 

Are your git skills sufficient to undo undo the squash, and undo my commit, and then just change the _str to _new?

rany2:
> Actually I see what you mean now.
> 
> Maybe the old formulation was indeed better, but with `new_shaper_rate_kbps` rather than `shaper_rate_kbps_str`.
> 
> Are your git skills sufficient to undo the squash, and undo my commit, and then just change the `shaper_rate_kbps_str` to `new_shaper_rate_kbps`?

Sure

lynxthecat:
Sorry!

rany2:
> Sorry!

No worries

lynxthecat:
Nice work and runs fine for me. Many thanks.

PR #208, opened 2023-07-04
Validate user configuration entries

### LINKED ISSUE #207: Config file validation
Some kind of validation of the config file provided by the user would be great, in that if a config is found to contain some invalid assignment, the script will abort giving the user the chance to figure out what changed in the new version and update their config properly.

rany2:
Closes #207

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -1719,7 +1719,7 @@ do
 done
 if ((invalid_config))
 then
-	log_msg "ERROR" "The config file: '${config_path}' contains errors. Please address them. Exiting now."
+	log_msg "ERROR" "The config file: '${config_path}' contains one or more errors. Exiting now."
 	exit 1
 fi
 unset invalid_config
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -1704,7 +1704,17 @@ typeof() {
 }
 
 validate_config_entry() {
-	# Must be called before loading config_path into the global scope
+	# Must be called before loading config_path into the global scope.
+	#
+	# When the entry is valid, two types are returned with the first type
+	# being the invalid user type and second type is the default type with
+	# the user needing to adapt the config file so that the entry uses the
+	# default type.
+	#
+	# When the entry is valid, one type is returned and it will be the
+	# the type of either the default or user type. However because in that
+	# case they are both valid. It doesn't matter as they'd both have the
+	# same type.
 
 	local config_path="${1}"
 
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -1673,6 +1673,52 @@ debug_cmd()
 	fi
 }
 
+str_type() {
+	local -n str="${1}"
+
+	if [[ "${str}" =~ ^[0-9]+$ ]]
+	then
+		printf "number"
+	elif [[ "${str}" =~ ^[0-9]+\.[0-9]+$ ]]
+	then
+		printf "float"
+	else
+		printf "string"
+	fi
+}
+
+typeof() {
+	local type_sig=$(declare -p "${1}" 2>/dev/null)
+    if [[ "${type_sig}" =~ "declare --" ]]
+	then
+        str_type "${1}"
+    elif [[ "${type_sig}" =~ "declare -a" ]]
+	then
+        printf "array"
+    elif [[ "${type_sig}" =~ "declare -A" ]]
+	then
+        printf "map"
+    else
+        printf "none"
+    fi
+}
+
+validate_config_entry() {
+	# Must be called before loading config_path into the global scope
+
+	local config_path="${1}"
+
+	local user_type=$(. "${config_path}" && typeof "${2}")
+	local valid_type=$(typeof "${2}")
+
+	if [[ "${user_type}" != "${valid_type}" ]]
+	then
+		printf "${user_type} ${valid_type}"
+	else
+		printf "${valid_type}"
+	fi
+}
+
 # ======= Start of the Main Routine ========
 
 [[ -t 1 ]] && terminal=1 || terminal=0
@@ -1715,6 +1761,13 @@ do
 	then
 		invalid_config=1
 		log_msg "ERROR" "The value: '${value}' in config file: '${config_path}' is not a valid config entry."
+	else
+		read -r user supposed <<< $(validate_config_entry "${config_path}" "${value}")
+		if [[ -n "${supposed}" ]]
+		then
+			log_msg "ERROR" "The value: '${value}' in config file: '${config_path}' is not a valid ${supposed} value."
+			invalid_config=1
+		fi
 	fi
 done
 if ((invalid_config))
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -1769,14 +1769,17 @@ do
 			error_msg="The value of '${key}' in config file: '${config_path}' is not a valid ${supposed} value."
 
 			case "${user}" in
-				*float) error_msg="${error_msg} Also, floats are not supported." ;;
+				*float)          error_msg="${error_msg} Also, floats are not supported." ;;
 				negative-number) error_msg="${error_msg} Also, negative numbers are not supported." ;;
 				*) ;;
 			esac
 
 			log_msg "ERROR" "${error_msg}"
+			unset error_msg
+
 			invalid_config=1
 		fi
+		unset user supposed
 	fi
 done
 if ((invalid_config))
@@ -1784,7 +1787,7 @@ then
 	log_msg "ERROR" "The config file: '${config_path}' contains one or more errors. Exiting now."
 	exit 1
 fi
-unset invalid_config
+unset valid_config_entries user_config invalid_config key
 
 # shellcheck source=config.primary.sh
 . "${config_path}"
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -1769,8 +1769,7 @@ do
 			error_msg="The value of '${key}' in config file: '${config_path}' is not a valid ${supposed} value."
 
 			case "${user}" in
-				*float)          error_msg="${error_msg} Also, floats are not supported." ;;
-				negative-number) error_msg="${error_msg} Also, negative numbers are not supported." ;;
+				negative-*) error_msg="${error_msg} Also, negative numbers are not supported." ;;
 				*) ;;
 			esac
 
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -64,6 +64,8 @@ PREFIX=/root/cake-autorate
 . "${PREFIX}/lib.sh"
 # shellcheck source=defaults.sh
 . "${PREFIX}/defaults.sh"
+# get valid config overrides
+mapfile -t valid_config_entries < <(grep -E '^[^(#| )].*=' "${PREFIX}/defaults.sh" | sed -e 's/[\t ]*\#.*//g' -e 's/=.*//g')
 
 trap cleanup_and_killall INT TERM EXIT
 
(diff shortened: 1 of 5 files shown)


rany2:
@lynxthecat Thoughts on this?

lynxthecat:
Might it be worth wrapping up the config file validation into a function?

rany2:
It's never getting reused though, seems unnecessary

rany2:
Might be worth changing some of the errors returned for config parsing failures, like `log_msg "ERROR" "Config file error. Please check config file entries."`

lynxthecat:
I like your changes to cleanup_and_killall() in respect of the set +e and intercept_stderr. I didn't think of that issue.

rany2:
@lynxthecat does a config_file_check still make sense?

lynxthecat:
Ah that's clever - so you're checking all config value sets in the config against corresponding entries in defaults.sh to ensure any provided option is a valid option?

cake-autorate.sh
```diff
@@ -1688,6 +1699,26 @@ then
...
+	# shellcheck disable=SC2076
+	if [[ ! " ${valid_config_entries[*]} " =~ " ${value} " ]]
+	then
+		invalid_config=1
+		log_msg "ERROR" "${value} is an invalid config entry."
+	fi
+done
+if ((invalid_config))
+then
+	log_msg "ERROR" "Your config contains errors, please address them. Exiting now."
```
lynxthecat, line 1717: How about: "The config file: '${config_path}' contains errors. Please address them Exiting now."

cake-autorate.sh
```diff
@@ -1688,6 +1699,26 @@ then
...
+mapfile -t user_config < <(grep -E '^[^(#| )].*=' "${config_path}" | sed -e 's/[\t ]*\#.*//g' -e 's/=.*//g')
+invalid_config=0
+for value in "${user_config[@]}"
+do
+	[[ "${value}" == "config_file_check" ]] && continue
+	# shellcheck disable=SC2076
+	if [[ ! " ${valid_config_entries[*]} " =~ " ${value} " ]]
+	then
+		invalid_config=1
+		log_msg "ERROR" "${value} is an invalid config entry."
```
lynxthecat, line 1712: How about: "The value: '${value}' in config file: '${config_path}' is not a valid config entry."

rany2:
> Ah that's clever - so you're checking all config value sets in the config against corresponding entries in defaults.sh to ensure any provided option is a valid option?

Pretty much

rany2:
> @lynxthecat does the config_file_check variable still make sense?

P.S. do you think it should be removed? It doesn't really make sense IMO. We have a defaults.sh that ensures that all variables the script relies on are initialized.

lynxthecat:
Ah yes - seems to make sense to remove it.

cake-autorate.sh
```diff
@@ -1688,14 +1699,33 @@ then
...
+	if [[ ! " ${valid_config_entries[*]} " =~ " ${value} " ]]
+	then
+		invalid_config=1
+		log_msg "ERROR" "The value: '${value}' in config file: '${config_path}' is not a valid config entry."
+	fi
+done
+if ((invalid_config))
 then
-	log_msg "ERROR" "Config file error. Please check config file entries."
+	log_msg "ERROR" "The config file: '${config_path}' contains errors. Please address them. Exiting now."
```
lynxthecat, line 1722:
```suggestion
	log_msg "ERROR" "The config file: '${config_path}' contains one or more errors. Exiting now."
```

lynxthecat:
So far so good:

```
root@OpenWrt-1:~/cake-autorate# ./cake-autorate.sh
ERROR; 2023-07-04-16:30:25; 1688484625.526007; The value: 'woke_factor' in config file: '/root/cake-autorate/config.primary.sh' is not a valid config entry.
ERROR; 2023-07-04-16:30:25; 1688484625.530927; The config file: '/root/cake-autorate/config.primary.sh' contains errors. Please address them. Exiting now.
DEBUG; 2023-07-04-16:30:25; 1688484625.535466; Starting: cleanup_and_killall with PID: 7278
INFO; 2023-07-04-16:30:25; 1688484625.536791; Stopping cake-autorate with PID: 7278 and config: /root/cake-autorate/config.primary.sh
INFO; 2023-07-04-16:30:25; 1688484625.538115; Killing all background processes and cleaning up temporary files.
SYSLOG; 2023-07-04-16:30:26; 1688484626.546020; Stopped cake-autorate with PID: 7278 and config: /root/cake-autorate/config.primary.sh
```

lynxthecat:
@rany2 should we check that config values save for `ping_prefix_str` and `ping_extra_args` are set to positive integers?

rany2:
@lynxthecat You could test in this state now

rany2:
Things you might think are missing but are actually not: 

- Verify that a number is actually octal or not (Bash does not consider a 0 being preceded in a number to be octal)
- Better float regex, technically Bash does not have the concept of a float. It's only there so that were could in future warn user than floats are not supported and provide better errors. 

Things that are actually missing:

- Validation of maps/arrays. The type of the array content itself is not validated, however the fact that it is a map/array is considered regardless.
- I can't think of anything else off the top of my head right now.

rany2:
```
ERROR; 2023-07-04-22:02:52; 1688508172.051024; The value: 'dl_if' in config file: '/root/cake-autorate/config.primary.sh' is not a valid string value. Also, negative numbers are not supported.
ERROR; 2023-07-04-22:02:52; 1688508172.156770; The value: 'min_dl_shaper_rate_kbps' in config file: '/root/cake-autorate/config.primary.sh' is not a valid number value.
ERROR; 2023-07-04-22:02:52; 1688508172.199717; The config file: '/root/cake-autorate/config.primary.sh' contains one or more errors. Exiting now.
DEBUG; 2023-07-04-22:02:52; 1688508172.203309; Starting: cleanup_and_killall with PID: 51950
INFO; 2023-07-04-22:02:52; 1688508172.203607; Stopping cake-autorate with PID: 51950 and config: /root/cake-autorate/config.primary.sh
INFO; 2023-07-04-22:02:52; 1688508172.203869; Killing all background processes and cleaning up temporary files.
SYSLOG; 2023-07-04-22:02:53; 1688508173.208124; Stopped cake-autorate with PID: 51950 and config: /root/cake-autorate/config.primary.sh
Killing all instances of cake one-by-one now.
```

Could we get rid of the unix time stamp from the output and reduce the debug level by default?  It looks really cluttered with the default configuration.

lynxthecat:
First of all - wow. Let me catch up on this!

> Could we get rid of the unix time stamp from the output and reduce the debug level by default? It looks really cluttered with the default configuration. It seems difficult to read when an error occurs.

I agree with you that the output of cake-autorate looks a little cluttered. For a while I implemented padding, but once the columns grew to a certain extent it no longer helped so much. A consideration is that we have a couple of external plotting programs (e.g. [this](https://github.com/lynxthecat/cake-autorate/blob/master/fn_parse_autorate_log.m) and [this super cool prometheus exporter](https://github.com/bairhys/prometheus-cake-autorate-exporter/) that @bairhys wrote) that expect a certain format. I'm reluctant to remove the timestamp because so often timings are critical when checking control aspects like the bufferbloat refractory index. Reducing the default debug level has the drawback that when a user first tries cake-autorate and it does not work then they will not know why. In the end the output is only normally something you look at following problems or for offline analysis of data in a spreadsheet or one of the plotting tools. 

Any thoughts on the above?

lynxthecat:
Your typeof() is super cool:

```
root@OpenWrt-1:~# x=()
root@OpenWrt-1:~# printf "$(typeof x) \n"
array
root@OpenWrt-1:~# x=0
root@OpenWrt-1:~# printf "$(typeof x) \n"
array
root@OpenWrt-1:~# unset x
root@OpenWrt-1:~# x=0
root@OpenWrt-1:~# printf "$(typeof x) \n"
number
root@OpenWrt-1:~# x=0.2
root@OpenWrt-1:~# printf "$(typeof x) \n"
float
root@OpenWrt-1:~# x="string"
root@OpenWrt-1:~# printf "$(typeof x) \n"
string
root@OpenWrt-1:~# declare -A x
root@OpenWrt-1:~# printf "$(typeof x) \n"
map
```

lynxthecat:
OK here is one issue.

The defaults for dl_delay_thr_ms are whole numbers:

```
# delay threshold in ms is the extent of OWD increase to classify as a delay
# these are automatically adjusted based on maximum on the wire packet size
# (adjustment significant at sub 12Mbit/s rates, else negligible)
dl_delay_thr_ms=30 # (milliseconds)
ul_delay_thr_ms=30 # (milliseconds)
```

But actually these can be integers or floats. In my config I have:

```
dl_delay_thr_ms=40.0 # (milliseconds)
ul_delay_thr_ms=40.0 # (milliseconds)
```

I see this:

```
ERROR; 2023-07-05-09:17:53; 1688545073.297637; The value of 'dl_delay_thr_ms' in config file: '/root/cake-autorate/config.primary.sh' is not a valid number value.
ERROR; 2023-07-05-09:17:53; 1688545073.317344; The value of 'ul_delay_thr_ms' in config file: '/root/cake-autorate/config.primary.sh' is not a valid number value. Also, floats are not supported.
ERROR; 2023-07-05-09:17:53; 1688545073.338122; The value of 'dl_delay_thr_ms' in config file: '/root/cake-autorate/config.primary.sh' is not a valid number value.
```

lynxthecat:
@rany2 one issue is that with some of the configuration elements can be set to either a number or a float, and that mismatch can give unnecessary errors. But perhaps the way round this is just to ensure that the defaults are floats where floats are supported.

cake-autorate.sh
```diff
@@ -1688,14 +1745,52 @@ then
...
+		log_msg "ERROR" "The key: '${key}' in config file: '${config_path}' is not a valid config entry."
+	else
+		# shellcheck disable=SC2311
+		read -r user supposed <<< "$(validate_config_entry "${config_path}" "${key}")"
+		if [[ -n "${supposed}" ]]
+		then
+			error_msg="The value of '${key}' in config file: '${config_path}' is not a valid ${supposed} value."
+
+			case "${user}" in
+				*float)          error_msg="${error_msg} Also, floats are not supported." ;;
```
lynxthecat, line 1772: But we use floats like in (dl/ul)_delay_thr_ms?

rany2: They aren't?

rany2: Oh, I see..

lynxthecat:
Looks good. I'll retest shortly. This is a fantastic addition.

cake-autorate.sh
```diff
@@ -1688,14 +1745,51 @@ then
...
+	if [[ ! " ${valid_config_entries[*]} " =~ " ${key} " ]]
+	then
+		((config_error_count++))
+		log_msg "ERROR" "The key: '${key}' in config file: '${config_path}' is not a valid config entry."
+	else
+		# shellcheck disable=SC2311
+		read -r user supposed <<< "$(validate_config_entry "${config_path}" "${key}")"
+		if [[ -n "${supposed}" ]]
+		then
+			error_msg="The value of '${key}' in config file: '${config_path}' is not a valid ${supposed} value."
```
lynxthecat, line 1769:
```suggestion
			error_msg="The value of '${key}' in config file: '${config_path}' is not a valid value of type: ${supposed}."
```

lib.sh
```diff
@@ -16,6 +16,49 @@ then
...
+}
+
+str_type() {
+	# str_type -- returns the type of a string
+
+	local -n str="${1}"
+
+	if [[ "${str}" =~ ^[0-9]+$ ]]
+	then
+		printf "number"
```
lynxthecat, line 45: How about "positive-integer"?

lynxthecat:
All seems good. Should we handle entries that are not assignments like the word 'exit' or is that completely unecessary?

lynxthecat:
Is the unset to save memory?

rany2:
> Should we handle entries that are not comments or assignments like the word 'exit' or is that completely unnecessary?

Against that because users might want to define functions that could be used to assign to the variables; and if someone defined something like exit. It would be obvious something is broken. This is mostly to prevent issues where cake-autorate silently starts behaving improperly but doesn't provide feedback that it is because of the config is broken

lib.sh
```diff
@@ -16,6 +16,49 @@ then
...
+		printf "none"
+	fi
+}
+
+str_type() {
+	# str_type -- returns the type of a string
+
+	local -n str="${1}"
+
+	if [[ "${str}" =~ ^[0-9]+$ ]]
```
lynxthecat, line 43: Is it not possible in bash to make a case out of this? Or is the level of regular expression permitted in case statements not sufficient?

lynxthecat:
I think we're good to merge all of this. I can't think of anything else now. What do you think?

rany2:
@lynxthecat Guess so, go for it. We could later add better validation for arrays; should be simple.

rany2:
> What's the reason for the unset?

Consider this, sorry I didn't respond sooner. My email is really cluttered.

```
$ x=()
$ declare -p x
declare -a x=()
$ x=0
$ declare -p x
declare -a x=([0]="0")
# unset x; x=0
# declare -p x
declare -- x="0"
```

lynxthecat:
Thanks @rany2. Could I run something past you on IRC?

PR #224, opened 2023-07-05
Ensure setup.sh fetches the latest version of itself

rany2:


rany2:
@lynxthecat Please test this out, but I think it should work.

setup.sh
```diff
@@ -37,6 +37,13 @@ main() {
 	SRC_DIR="https://github.com/${REPOSITORY}/archive/"
 	API_URL="https://api.github.com/repos/${REPOSITORY}/commits/${BRANCH}"
 	DOC_URL="https://github.com/${REPOSITORY}/tree/${BRANCH}#installation-on-openwrt"
+	CURRENT_SETUP_SH="https://raw.githubusercontent.com/${REPOSITORY}/${BRANCH}/setup.sh"
+
+	# Get the current setup.sh script and execute it
+	if [ -z "${__CAKE_AUTORATE_URL_LOADED:-}" ]
```
lynxthecat, line 43: What does this do?

rany2: When the setup.sh file is run, it fetches the branch's latest setup.sh from the web and then executes it; but it needs to figure out whether it should keep fetching from the latest or not; so `__CAKE_AUTORATE_URL_LOADED` is a marker for the script to know that it shouldn't fetch itself.

rany2: Otherwise you'd have an infinite loop of it just constantly loading itself up from the web and executing it.

rany2:
Don't merge it yet, still need to preserve arguments.

lynxthecat:
Should we log downloading new script and latest commit / updating to commit number?

rany2:
> Should we log downloading new script and latest commit / updating to commit number?

I don't follow

lynxthecat:
If we overwrite setup.sh should we not log that? Also should we not inform user of any upgrade? So print out the commit number previously installed and the one that has just been installed?

rany2:
It's not actually overwriting anyway, it's just loading it on the fly

lynxthecat:
OK but what about informing user of the upgrade.. e.g. upgraded from version/commit to version/commit?

rany2:
It's not knowable anyway, we can't do much here.

rany2:
It's unconditionally going to load it from github if `__CAKE_AUTORATE_URL_LOADED` was not set

lynxthecat:
We know previous version the version file you created. And we know current version because you obtain that with the setup.sh already. 

By the way when running I see:

```
root@OpenWrt-1:~/cake-autorate# chmod +x setup.sh
root@OpenWrt-1:~/cake-autorate# ./setup.sh
Installing cake-autorate in /root/cake-autorate...
```

rany2:
Meh, too complicated.

PR #243, opened 2023-07-13
Fix change to README that broke the maint script

rany2:


richb-hanover:
Perhaps this HTML comment should precede the paragraph that I broke :-)
```
<!-- The following sentence must wrap the version in <span id="version">x.y.z</span> tags -->
```
It might need to go in both `master` and `v3.0` branches...

PR #249, opened 2023-08-17
Use comma as delimiter for tsping

### LINKED ISSUE #246: tsping mode is not compatible with mwan3
Some time ago, the code was changed to pass `--machine-readable=' '` argument to tsping. Unfortunately, the `mwan3 use lte exec` wrapper (prefix command) mangles this. It is a bug in `mwan3`, but bugs there are never fixed. So please try to avoid triggering it. In other words, please try to avoid the need for any special characters that need quoting or otherwise cannot pass through the `$*` construct, such as spaces. Maybe use $IFS that includes the comma when parsing.

rany2:
Using "mwan3 use X ..." for tsping does not work properly due to the developers calling "..." using $*. This works around the issue by using `,` as the delimiter instead.

Fixes #246

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -28,6 +28,9 @@ cake_autorate_version="3.1.0-PRERELEASE"
 ## accessible via fds in the form: ${process_name_fd}
 ## thereby to enable transferring instructions and data between processes
 
+# Set the IFS to space and comma
+IFS=" ,"
+
 # Initialize file descriptors
 ## -1 signifies that the log file fd will not be used and
 ## that the log file will be written to directly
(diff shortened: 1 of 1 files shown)


lynxthecat:
Unfortunately this breaks tsping - no reflector responses are read in.

rany2:
@lynxthecat I could see responses on my end, at any rate could you try now?

lynxthecat:
Still not working.

lynxthecat:
It works with:

```
 565 parse_preprocessor()
 566 {
 567         # prepend REFLECTOR_RESPONSE and append timestamp as a checksum
 568         while IFS=', ' read -r timestamp remainder
```

Won't this issue affect the other ping binaries too?

PR #257, opened 2023-10-31
fix: Delay startup before assigning initial timestamps

joshuawn:
When startup_wait_s is non-zero, anything dependent upon initial timestamps (e.g., the reflector timeout period) will be affected by the sleep. If the reflector timeout period is less than the startup_wait_s duration. this causes a fatal crash immediately after the startup sleep ends.
This is addressed by sleeping before assigning any initial timestamps.

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -2036,6 +2036,13 @@ set_shaper_rate "ul"
 
 update_max_wire_packet_compensation
 
+# Wait if ${startup_wait_s} > 0
+if ((startup_wait_us>0))
+then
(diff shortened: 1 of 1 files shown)


lynxthecat:
Seems reasonable. Exactly which variable is not initialized in this scenario. Is it somewhere around here:

https://github.com/lynxthecat/cake-autorate/blob/ef4600cde23a927ce4b8265ac510dffab4166854/cake-autorate.sh#L2251C29-L2251C69

lynxthecat:
@joshuawn many thanks indeed for submitting this pull request. I was able to recreate the issue and your proposed fix resolved it. And otherwise I agree with where you have relocated the startup delay code.

PR #272, opened 2024-02-08
Increase idle thr & improve config documentation

lynxthecat:
The default idle threshold of 1MBit/s seems to be too low for a typical set of home IoT devices, resulting in the sleep activation logic never triggering  - see [here](https://forum.openwrt.org/t/cake-w-adaptive-bandwidth/135379/3507). In any case, it seems apt to add this as one of the configuration options present in config.primary.sh.

And the term 'instance' is more appropriate than 'interface' in respect of config.x.sh.

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -1816,7 +1816,7 @@ then
 		broken_log_file_path_override=${log_file_path_override}
 		log_file_path=/var/log/cake-autorate${instance_id:+.${instance_id}}.log
 		log_msg "ERROR" "Log file path override: '${broken_log_file_path_override}' does not exist. Exiting now."
-		exit
+		exit 1
 	fi
 	log_file_path=${log_file_path_override}/cake-autorate${instance_id:+.${instance_id}}.log
 else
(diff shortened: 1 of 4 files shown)


rany2:
I don't like defaulting this to 2Mbit/s because of ADSL. If we go forward with this I think there needs to be a sanity check that makes sure that the connection_active_thr_kbps is lower than either min dl/ul shaper rate.

If the check is added then at least ADSL users would notice this misconfiguration.

lynxthecat:
Good point. I added the sanity check.

cake-autorate.sh
```diff
@@ -1867,6 +1867,9 @@ command -v "${pinger_binary}" &> /dev/null || { log_msg "ERROR" "ping binary ${p
 # Check bufferbloat detection threshold not greater than window length
 (( bufferbloat_detection_thr > bufferbloat_detection_window )) && { log_msg "ERROR" "bufferbloat_detection_thr cannot be greater than bufferbloat_detection_window. Exiting script."; exit; }
 
+(( connection_active_thr_kbps > min_dl_shaper_rate_kbps )) && { log_msg "ERROR" "connection_active_thr_kbps cannot be greater than min_dl_shaper_rate_kbps. Exiting script."; exit; }
```
rany2, line 1870: Have it exit with error, so `exit 1`

rany2 (approved):

PR #276, opened 2024-04-03
Gracefully deal with incomplete lines in the logfile

moeller0:
On encountering an incomplete line, stop parsing, and work with the complete records parsed so far.

### DIFF
--- a/fn_parse_autorate_log.m
+++ b/fn_parse_autorate_log.m
@@ -850,6 +850,11 @@ function [ autorate_log, log_FQN ] = fn_parse_autorate_logfile( log_FQN, command
 			disp([num2str(line_count), ': ', current_line]);
 		endif
 
+    if (length(current_line) < 5)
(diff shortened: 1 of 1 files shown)


rany2 (approved):
Thanks @moeller0

lynxthecat:
@moeller0 how about skipping any bad line and then continuing on with any remaining lines?

rany2:
@lynxthecat your log should never be broken like this in the first place, I think it's appropriate for it to fail. Making this a warning will just mean that it will be ignored and useless.

lynxthecat:
Also just wondering for those that mess about with tail and merging lines etc.

moeller0:
I am more inclined to let users that break their log files like that keep both pieces...

It is one thing to not crash on a problematic logfile, but a whole different game trying to deduce what to trust in a non conforming logfile...
As they say, trying to make things fool proof just creates selective pressure to evolve 'better' fools ;)

PR #277, opened 2024-04-06
Robustify the log parser and add the SUMMARY record type

moeller0:
Teach the parser to skip over lines not containing the expected number of fields.
Also add the SUMMARY record type to the parser to clean up the output.

As it says on the tin, make it more robust and less verbose... Currently SUMMARY records are fully parsed, but not used any further...

### DIFF
--- a/fn_parse_autorate_log.m
+++ b/fn_parse_autorate_log.m
@@ -186,7 +186,7 @@ function [ ] = fn_parse_autorate_log( log_FQN, plot_FQN, x_range_sec, selected_r
 		min_sequence_number = 1;
 
 		align_rate_and_delay_zeros = 1; % so that delay and rate 0s are aligned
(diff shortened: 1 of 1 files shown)


fn_parse_autorate_log.m
```diff
@@ -787,6 +787,9 @@
 	log_struct.INFO = [];
 	log_struct.SHAPER = [];
 	log_struct.metainformation = [];
+  log_struct.SUMMARY_HEADER = [];
+	log_struct.SUMMAY = [];
```
rany2, line 791: This looks like a typo, I could only see `log_struct.SUMMARY` being used.

moeller0: Indeed, good catch, thanks...

rany2:
Thanks a lot @moeller0 , merged!

PR #280, opened 2024-04-06
Actually store the SUMMARY data to the output struct

moeller0:
While we are not using this right now, no reason to not save it to the mat file. Also fix a comment up...

The previous changes fixed SUMMARY parsing, but this now actually stores the SUMMARY data to the matgile and makes it available.

### DIFF
--- a/fn_parse_autorate_log.m
+++ b/fn_parse_autorate_log.m
@@ -1068,7 +1068,7 @@ function [ cur_record_type ] = fn_get_record_type_4_line( current_line, delimite
       elseif strcmp(current_line(1:8), "SUMMARY;")
         cur_record_type = "SUMMARY";
         if ~isfield(log_struct.metainformation, 'SUMMARY_HEADER') || log_struct.metainformation.SUMMARY_HEADER.count < 1
-          # we have not encountered a SUMMARY_HEADER record yet and do not know how to parse REFLECTOR records, so SKIP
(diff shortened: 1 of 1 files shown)


fn_parse_autorate_log.m
```diff
@@ -1220,7 +1220,7 @@
 function [ ] = fn_parse_current_line( cur_record_type, current_line, delimiter_string, line_increment)
 global log_struct
 
-if ~ismember(cur_record_type, {'INFO', 'SHAPER', 'DEBUG', 'DATA', 'LOAD', 'REFLECTOR'}) % {'DEBUG', 'INFO', 'SJHAPER'}
```
rany2, line 1223: Might as well fix `SJHAPER => SHAPER`

rany2:
Are you opposed to running the code through a formatter? There's some inconsistent indentation

moeller0:
Well, I appreciate code review, but talking about white space is IMHO mostly busy work (I am not opposed to some level of consistency, but in the grand scheme of things, I consider this to be cosmetic issues)... That said it appears on my new machine I need to change the editor defaults to my previous defaults, I assume we mostly talking spaces versus tabs?

rany2:
I don't disagree. It is just a minimum effort change to run it through a code reformatter, but if it's not causing any issues on your end then there is no need.

PR #281, opened 2024-04-06
Add an Octave formatter and make our script compliant with formatter

rany2:
## Add Octave formatting script

Modified from matlab-formatter-vscode to add GNU Octave support and to adapt
it to cake-autorate's needs (i.e., I added some features like inplace, tabs,
argparse, etc.).

Further, I fixed in which a line like:
`if ...'LOAD', 'REFLECTOR'}) % {'DEBUG', 'INFO', 'SJHAPER'}` would break the
formatter in addition to other more subtle issues.

It is based on https://github.com/affenwiesel/matlab-formatter-vscode commit 43d7224.

For reference on what needs to be modified, see
https://en.wikibooks.org/wiki/MATLAB_Programming/Differences_between_Octave_and_MATLAB

## [Octave] Avoid `end` and use `endif`/`endwhile` consistently

In some parts of the script, there is some usage of `end`.
It doesn't really matter but I prefer for the script to be
more consistent.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+cake-autorate.*.log
+cake-autorate.*.log.gz
+cake-autorate.*.log.mat
+output.*.*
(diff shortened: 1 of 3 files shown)


rany2:
**NOTE:** I didn't reformat the Octave script, @moeller0 could do that on his end whenever he feels like. I don't want to break any branches on his end as he's the primary developer of this. It's just nice to have the option to reformat if we ever needed to.

PR #283, opened 2024-04-06
Improve maintain log file

lynxthecat:


cake-autorate.sh
```diff
@@ -97,6 +97,8 @@ cleanup_and_killall()
 	fi
 	terminate "${proc_pids[@]}"
 
+	sleep_s 1
```
rany2, line 100: Why are you sleeping for 1sec here?

lynxthecat: Good point. I forgot to remove this. I've removed it now.

cake-autorate.sh
```diff
@@ -302,22 +302,21 @@ flush_log_pipe()
...
+	kill_maintain_log_file_signalled=1
+	log_msg "DEBUG" "Killing maintain_log_file."
 }
 
 maintain_log_file()
 {
 	trap '' INT
 	trap 'kill_maintain_log_file' TERM EXIT
-	trap 'export_log_file' USR1
+	trap 'export_log_file_signalled=1"' USR1
```
rany2, line 314: typo, should be `export_log_file_signalled=1` without `"`

cake-autorate.sh
```diff
@@ -346,10 +347,23 @@ maintain_log_file()
 				((log_file_size_KB=log_file_size_bytes/1024))
 				log_msg "DEBUG" "log file size: ${log_file_size_KB} KB has exceeded configured maximum: ${log_file_max_size_KB} KB so flushing and rotating log file."
 				break
-			elif (( reset_log_file_signalled ))
```
rany2, line 349: nitpicking but this is very awkward, instead of `elif (( kill_maintain_log_file_signalled || export_log_file_signalled || reset_log_file_signalled ))`... do:

```shell
				elif ((kill_maintain_log_file_signalled))
				then
					flush_log_pipe
					trap - TERM EXIT
					exit
				elif ((export_log_file_signalled))
				then
					log_msg "DEBUG" "received log file export signal so exporting log file."
					export_log_file
					export_log_file_signalled=0
				elif ((reset_log_file_signalled))
				then
					log_msg "DEBUG" "received log file reset signal so flushing and resetting log file."
					break
				fi
```

lynxthecat: I thought about that, but surely that would be a lot slower?

rany2: I don't know if opening a new arithmetic context is slower, but if you think it's slower then maybe you should avoid it in general and stick to `test`

cake-autorate.sh
```diff
@@ -328,11 +331,13 @@ maintain_log_file()
 
 		t_log_file_start_us=${EPOCHREALTIME/.}
 
-		while read -r -N "${log_file_buffer_size_B}" -u "${log_fd}" log_chunk
```
rany2, line 331: What was wrong with just adding `-t0` to peek? Also, I don't like how this code is structured:

```
while true
do
   while true
   do
````

PR #284, opened 2024-04-06
Fully commit on octave-first and octave-only for end statements

moeller0:
Matlab only accepts a generic "end" while octave allows to document scope better by having distinct end variant like:
if ... endif
for ... endfor

While the code originally started under matlab, by now it is essentially octave only anyway so convert the last unadorned ends...

### DIFF
--- a/fn_parse_autorate_log.m
+++ b/fn_parse_autorate_log.m
@@ -33,7 +33,7 @@ function [ ] = fn_parse_autorate_log( log_FQN, plot_FQN, x_range_sec, selected_r
 
 %if ~(isoctave)
 dbstop if error;
-%end
+%endif
 
 timestamps.(mfilename).start = tic;
 fq_mfilename = mfilename('fullpath');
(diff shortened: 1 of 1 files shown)


rany2:
I think I already made this change here: https://github.com/lynxthecat/cake-autorate/commit/68546ffb4cc03b6bbc0254c93cb6cfa42728d3d6 but I'll merge it anyway.

rany2 (approved):

rany2:
Yep, the two commits are in fact identical. It's pretty strange how it shows up in the git history though :)

rany2:
@moeller0 I'm not sure if this interests you but it might prove helpful to you too: https://github.com/lynxthecat/cake-autorate/tree/d943630cc1ffb8e653d5b540c9ee21889c4e3d11/maint#octave-formatter-guide

It should be helpful if you ever wanted to go through your changes using `git diff`.

moeller0:
I will take a "pass" on this; nevertheless thanks.

PR #287, opened 2024-04-07
Slight cleanup of maintain_log_file and log_msg

rany2:
Untested


rany2:
I'll make another PR with more changes that will include this in the future.

PR #290, opened 2024-04-10
Don't use inefficient arrays to queue signal in maintain_log

rany2:
CC @lynxthecat

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -300,10 +300,11 @@ flush_log_pipe()
 
 maintain_log_file()
 {
+	signal=0
 	trap '' INT
-	trap 'signal+=(KILL)' TERM EXIT
-	trap 'signal+=(EXPORT)' USR1
-	trap 'signal+=(RESET)' USR2
+	trap 'signal=$(hexlify_int $((15<<0 | 0x${signal})))' TERM EXIT  # KILL
+	trap 'signal=$(hexlify_int $((15<<4 | 0x${signal})))' USR1       # EXPORT
+	trap 'signal=$(hexlify_int $((15<<8 | 0x${signal})))' USR2       # RESET
 
 	log_msg "DEBUG" "Starting: ${FUNCNAME[0]} with PID: ${BASHPID}"
 
@@ -346,26 +347,29 @@ maintain_log_file()
 			fi
 
 			# Check for signals
-			case ${signal[0]-} in
-				KILL)
+			case ${signal} in
+				0)
+					;;
+				*f)
 					log_msg "DEBUG" "received log file kill signal so flushing log and exiting."
 					flush_log_pipe
 					trap - TERM EXIT
 					exit
 					;;
-				EXPORT)
+				*f0)
 					log_msg "DEBUG" "received log file export signal so exporting log file."
 					export_log_file
-					signal=("${signal[@]:1}")
+					signal=$(hexlify_int $((~(15<<4) & 0x${signal})))
 					;;
-				RESET)
+				f00)
 					log_msg "DEBUG" "received log file reset signal so flushing log and resetting log file."
 					flush_log_pipe
 					reset_log_file
-					signal=("${signal[@]:1}")
+					signal=$(hexlify_int $((~(15<<8) & 0x${signal})))
 					break
 					;;
 				*)
+					log_msg "ERROR" "received unknown signal: ${signal}."
 					;;
 			esac
 		done
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -301,14 +301,13 @@ flush_log_pipe()
 maintain_log_file()
 {
 	trap '' INT
-	trap 'signal=KILL' TERM EXIT
-	trap 'signal=EXPORT' USR1
-	trap 'signal=RESET' USR2
+	trap 'signal+=(KILL)' TERM EXIT
+	trap 'signal+=(EXPORT)' USR1
+	trap 'signal+=(RESET)' USR2
 
 	log_msg "DEBUG" "Starting: ${FUNCNAME[0]} with PID: ${BASHPID}"
 
 	printf -v log_file_buffer_timeout_s %.1f "${log_file_buffer_timeout_ms}e-3"
-	signal=""
 
 	while true
 	do
@@ -347,7 +346,7 @@ maintain_log_file()
 			fi
 
 			# Check for signals
-			case ${signal} in
+			case ${signal[0]-} in
 				KILL)
 					log_msg "DEBUG" "received log file kill signal so flushing log and exiting."
 					flush_log_pipe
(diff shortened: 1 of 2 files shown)


rany2:
@lynxthecat can you test if negatively impacts performance on your router compared to current master?

lynxthecat:
OK I'll test. But any loss with using arrays on signal interception (exception rather than the norm) is surely outweighed by the loss with using three tests without any signals (norm rather than the exception)?

rany2:
@lynxthecat it's just to get an idea, i'll have it use switch-cases in a bit

lynxthecat:
```
root@OpenWrt-1:~/cake-autorate# ./bench_cpu.sh
Running CPU benchmark for test period of: 60s.
...
Average CPU usage over test period of 60s was: 6.679%

root@OpenWrt-1:~/cake-autorate# cp /tmp/cake-autorate/cake-autorate.sh .
root@OpenWrt-1:~/cake-autorate# vim cake-autorate.sh
root@OpenWrt-1:~/cake-autorate# ./bench_cpu.sh
Running CPU benchmark for test period of: 60s.
...
Average CPU usage over test period of 60s was: 7.632%
```

rany2:
@lynxthecat test now please

lynxthecat:
OK, but isn't it getting more and more ugly!

rany2:
@lynxthecat thanks!

lynxthecat:
```
root@OpenWrt-1:~/cake-autorate# ./bench_cpu.sh
Running CPU benchmark for test period of: 60s.
...
Average CPU usage over test period of 60s was: 6.553%
```

rany2:
@lynxthecat did you use the variant that logs out "unknown signal 0"?

lynxthecat:
> @lynxthecat did you use the variant that logs out "unknown signal 0"?

How do you mean?

lynxthecat:
Oh I see, I'll retest.

lynxthecat:
```
root@OpenWrt-1:~/cake-autorate# cp /tmp/cake-autorate/cake-autorate.sh .
root@OpenWrt-1:~/cake-autorate# vim cake-autorate.sh
root@OpenWrt-1:~/cake-autorate# ./bench_cpu.sh
Running CPU benchmark for test period of: 60s.
...
Average CPU usage over test period of 60s was: 6.649
```

rany2:
What's the baseline from master?

lynxthecat:
Doing 2x head-to-head comparison. Two seconds.

lynxthecat:
Here are 2x runs from this pull request, then 2x runs from master:

```
root@OpenWrt-1:~/cake-autorate# ./bench_cpu.sh
Running CPU benchmark for test period of: 60s.
...
Average CPU usage over test period of 60s was: 6.649%
root@OpenWrt-1:~/cake-autorate# ./bench_cpu.sh
Running CPU benchmark for test period of: 60s.
...
Average CPU usage over test period of 60s was: 6.629%
root@OpenWrt-1:~/cake-autorate# cp /tmp/cake-autorate/cake-autorate.sh .
root@OpenWrt-1:~/cake-autorate# ./bench_cpu.sh
Running CPU benchmark for test period of: 60s.
...
Average CPU usage over test period of 60s was: 6.607%
root@OpenWrt-1:~/cake-autorate# ./bench_cpu.sh
Running CPU benchmark for test period of: 60s.
...
Average CPU usage over test period of 60s was: 6.602%
```

rany2:
I have another idea, just give me a moment @lynxthecat

rany2:
@lynxthecat please try the current solution

lynxthecat:
I will once I return from walking our dog!

lynxthecat:
Why not just put the printf calls inside the traps?

rany2:
@lynxthecat because that would just be too ugly!

lynxthecat:
Hmm! My array solution still seems preferable on the elegance score.

rany2:
> Hmm! My array solution still seems preferable on the elegance score.

I don't like your solution because it means an impatient user could spam cake-autorate with say an export request and cake-autorate will just mindlessly keep queueing it. This prevents this situation without any additional checks while allowing users to make, for example, both export and reset requests. Also no dynamic array sizes.

> So we are converting to hex with function call to provide string comparison with hex?

Yes, so we don't have to just match all the possible cases like we did before. I don't want to just do `signal & (1<<1)` and to be able to use a switch-case, I'd need to match all the possible cases with `1<<1` being set. So, `1<<1|1<<2` if `1<<2` was to be processed next in the switch-case order

lynxthecat:
There is surely an easier way. How about only adding array element if unique.

lynxthecat:
"${array[@]/$element/}" != "${array[@]}"

rany2:
> Also I’m not sure the export spam thing is so bad. I mean we’re honouring the requests sent by the user.

It is, because you also are waiting 0.5 seconds between every time you're processing it. So it could easily mean that the user will need to restart cake-autorate.

> There is surely an easier way. How about only adding array element if unique.
>
>"${array[@]/$element/}" != "${array[@]}"

I just felt like this solution was more efficient

lynxthecat:
> It is, because you also are waiting 0.5 seconds between every time you're processing it. So it could easily mean that the user will need to restart cake-autorate.

How do you mean?

rany2:
> How do you mean? I mean what’s the significance of the timeout. And that’ll only be if read doesn’t return sooner.

It makes the situation of queueing up these requests with your solution so much worse.

rany2:
Anyway how does this perform on your router, I'm  sure it's more efficient than the current implementation but by how much?

lynxthecat:
```
root@OpenWrt-1:~/cake-autorate# ./bench_cpu.sh
Running CPU benchmark for test period of: 60s.
...
Average CPU usage over test period of 60s was: 6.607%
root@OpenWrt-1:~/cake-autorate# ./bench_cpu.sh
Running CPU benchmark for test period of: 60s.
...
Average CPU usage over test period of 60s was: 6.578%
```

lynxthecat:
Same speedup can be given to master by just checking for null string:

```
root@OpenWrt-1:~/cake-autorate# ./bench_cpu.sh
Running CPU benchmark for test period of: 60s.
...
Average CPU usage over test period of 60s was: 6.599%
```

rany2:
Closing as the solution was too unreadable.

lynxthecat:
For the record, I admire the tenacity reflected in this approach. I am a huge fan of tenacity; it is surely a powerful force to be reckoned with.

rany2:
Thanks @lynxthecat, though I was pretty disappointed in myself when I saw how simple your solution was. I have no idea why my thinking was so skewed towards this, if this were in C the initial variant of this PR would have made a ton of sense but it just kept getting more absurd until we reached this final bit with converting it to hex... :rofl: 

So that was tragic !

rany2:
> until we reached this final bit with converting the thing to hex...

Oh, and how could I forget the masterpiece of just having a case contain all the possible combinations for a signal when I was just setting one bit at a time X)

PR #291, opened 2024-04-11
Some more bash optimizations

rany2:
The below changes reduces CPU usage by 0.5% on my X86 machine:

- Switch to using `:` instead of `true` as it is consistently faster in the below benchmark, unsure why as they are both noops:

  + timeout 3 bash -c 'while :; do true; echo y; done'
  + wc -l 1881683
  + timeout 3 bash -c 'while :; do :; echo y; done'
  + wc -l 2152695

- Switch to switch case for the idle/sleep function handling code.

- Remove unneded non-empty string handling check

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -70,7 +70,7 @@ cleanup_and_killall()
 	# Do not fail on error for this critical cleanup code
 	set +e
 
-	trap true INT TERM EXIT
+	trap : INT TERM EXIT
 	
 	log_msg "DEBUG" "Starting: ${FUNCNAME[0]} with PID: ${BASHPID}"
 	
@@ -196,7 +196,7 @@ rotate_log_file()
 
 	[[ -f ${log_file_path} ]] || return
 	cat "${log_file_path}" > "${log_file_path}.old"
-	true > "${log_file_path}"
+	: > "${log_file_path}"
 }
 
 reset_log_file()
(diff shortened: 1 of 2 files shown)


lynxthecat:
I also came across this: 

https://lobste.rs/s/1l1t8f/bash_functions_are_better_than_i_thought

With some functions like log_msg() there are conditionals that are repeatedly reevaluated. It might be beneficial to either (re)define the functions based on the conditionals on  the initial script load and/or leverage evaluations like ${function} in which the function is changed in dependence upon the conditionals.

rany2:
@lynxthecat I've thought of this actually but got lazy. I'm not sure how beneficial it would be to do this but I could look into it

lynxthecat:
I am seeing more frequent and more prolonged dips down to 6% total CPU usage now (as measured using `atop 2` and hitting `p`:

image

cake-autorate.sh
```diff
@@ -1455,38 +1448,40 @@ do
...
-							sustained_connection_idle=1 t_last_connection_idle_us=t_start_us
-						fi
-						if ((t_sustained_connection_idle_us > sustained_idle_sleep_thr_us))
-						then
-							change_state_main "IDLE"	
-
-							log_msg "DEBUG" "Connection idle. Waiting for minimum load."
-
-							if ((min_shaper_rates_enforcement))
+					case ${load_condition[dl]}${load_condition[ul]} in
```
lynxthecat, line 1451: Nice!

rany2: Felt clever coming up with that trick :D

PR #294, opened 2024-04-15
ASUSWrt Fixes for setup.sh

rany2:
CC @LiveWire1968

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -32,7 +32,9 @@ main() {
 
 	# Check if OS is OpenWRT or derivative
 	unset ID_LIKE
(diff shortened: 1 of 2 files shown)


lynxthecat:
@rany2 you are a legend for pushing this forward and making it happen! Asus Merlin users will be very happy since up until now they've fudged it by disabling cake then performing speed test to set the bandwidth, then re-enabling cake. Switching to cake-autorate is vastly superior.

lynxthecat:
@rany2 is this usable yet?

rany2:
Later

PR #297, opened 2024-04-28
Silly optimizations

rany2:


rany2:
a8f8b8a25371047d598ccd53658278c736920eca actually is slower by a bit, it makes using EPOCHREALTIME nicer but not worth the slight perf penalty.

PR #299, opened 2024-04-29
Add remaining Asus Merlin fixes to installer

lynxthecat:
@LiveWire1968 any chance you could test the latest code on this asus-merlin branch and we'll add any necessary remaining fixes?

I think(?) the installation instructions should be something like:

Make sure these are installed: entware; bash; iputils-ping; and fping.

And then run:

```
wget https://raw.githubusercontent.com/lynxthecat/cake-autorate/asus-merlin/setup.sh
chmod +x ./setup.sh
./setup.sh lynxthecat/cake-autorate asus-merlin
```

And then edit the config file to specify the correct interfaces:

```
dl_if=ifb4eth0 # download interface
ul_if=eth0     # upload interface
```
and bandwidths (change according to your connection specifics):

```
min_dl_shaper_rate_kbps=5000   # minimum bandwidth for download (Kbit/s)
base_dl_shaper_rate_kbps=60000 # steady state bandwidth for download (Kbit/s)
max_dl_shaper_rate_kbps=100000 # maximum bandwidth for download (Kbit/s)

min_ul_shaper_rate_kbps=20000  # minimum bandwidth for upload (Kbit/s)
base_ul_shaper_rate_kbps=30000 # steady state bandwidth for upload (KBit/s)
max_ul_shaper_rate_kbps=35000  # maximum bandwidth for upload (Kbit/s)[/CODE]
```

And also enable logging:

```
output_summary_stats=1    # enable (1) or disable (0) output monitoring lines showing summary stats
```

And then try running manually:

```
./cake-autorate.sh
```

Does it work?

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -14,7 +14,7 @@ main() {
 	set -eu
 
 	# Setup dependencies to check for
-	DEPENDENCIES="jsonfilter wget tar grep"
+	DEPENDENCIES="jsonfilter wget tar grep bash"
 
 	# Set up remote locations and branch
 	REPOSITORY=${CAKE_AUTORATE_REPO:-${1-lynxthecat/cake-autorate}}
@@ -80,16 +80,6 @@ main() {
 		fi
 	done
 	[ "${exit_now}" -ge 1 ] && exit "${exit_now}"
-
-	# Retrieve required packages if not present
-	# shellcheck disable=SC2312
-	if [ "$(opkg list-installed | grep -Ec '^(bash|fping) ')" -ne 2 ]
-	then
-		printf "Running opkg update to update package lists:\n"
-		opkg update
-		printf "Installing bash and fping packages:\n"
-		opkg install bash fping
-	fi
 
 	# Create the cake-autorate directory if it does not exist
 	mkdir -p "${SCRIPT_PREFIX}" "${CONFIG_PREFIX}"
@@ -171,9 +161,12 @@ main() {
 	sed -e "s|%%SCRIPT_PREFIX%%|${SCRIPT_PREFIX}|g" -e "s|%%CONFIG_PREFIX%%|${CONFIG_PREFIX}|g" \
 		"${tmp}/launcher.sh.template" > "${SCRIPT_PREFIX}/launcher.sh"
 
-	# Also generate the service file from cake-autorate.template but DO NOT ACTIVATE IT
-	sed "s|%%SCRIPT_PREFIX%%|${SCRIPT_PREFIX}|g" "${tmp}/cake-autorate.template" > /etc/init.d/cake-autorate
-	chmod +x /etc/init.d/cake-autorate
+	# Also for OpenWrt generate the service file from cake-autorate.template but DO NOT ACTIVATE IT
+	if [ "${MY_OS}" = "openwrt" ]
+	then
+		sed "s|%%SCRIPT_PREFIX%%|${SCRIPT_PREFIX}|g" "${tmp}/cake-autorate.template" > /etc/init.d/cake-autorate
+		chmod +x /etc/init.d/cake-autorate
+	fi
 
 	# Get version and generate a file containing version information
 	cd "${SCRIPT_PREFIX}"
@@ -190,8 +183,20 @@ main() {
 	printf '\n%s\n\n' "${version} successfully installed, but not yet running"
 	printf '%s\n' "Start the software manually with:"
 	printf '%s\n' "   cd ${SCRIPT_PREFIX}; ./cake-autorate.sh"
-	printf '%s\n' "Run as a service with:"
-	printf '%s\n\n' "   service cake-autorate enable; service cake-autorate start"
+
+	case "${MY_OS}" in
+
+		"openwrt")
+			printf '%s\n' "Run as a service with:"
+			printf '%s\n\n' "   service cake-autorate enable; service cake-autorate start"
+			;;
+		"asuswrt")
+			printf '%s\n' "Launch script on boot with:"
+			printf '%s\n\n' "   echo ${SCRIPT_PREFIX}/launcher.sh >> /jffs/scripts/post-mount"
+			;;
+		*)
+			;;
+	esac
 }
 
 # Now that we are sure all code is loaded, we could execute the function
--- a/setup.sh
+++ b/setup.sh
@@ -85,6 +85,12 @@ main() {
 	done
 	[ "${exit_now}" -ge 1 ] && exit "${exit_now}"
 
+	if ! type "fping" >/dev/null 2>&1; then
+		printf "Warning, fping is required by default, but it is not installed.\n"
+		pritnf "So cake-autorate will not run with default settings.\n"
+		printf "Either install fping or ensure that one of the other supported ping binaries is installed and selected in the config(s).\n"
+	fi
+
 	# Create the cake-autorate directory if it does not exist
 	mkdir -p "${SCRIPT_PREFIX}" "${CONFIG_PREFIX}"
 
--- a/setup.sh
+++ b/setup.sh
@@ -32,9 +32,13 @@ main() {
 
 	# Check if OS is OpenWRT or derivative
 	unset ID_LIKE
-	# We do `set +/-e` here because in some Busybox sh versions
-	# `. /not_found || true` doesn't do anything
-	set +e; . /etc/os-release 2>/dev/null; set -e
+
+	# The following formulations seem to be problematic for some sh versions:
+	# - `. /not_found || true`
+	# - set +e; . /etc/os-release 2>/dev/null; set -e
+	# but this seems to work OK:
+	[ -e "/etc/os-release" ] && . /etc/os-release
+
 	for x in ${ID_LIKE:-}
 	do
 		if [ "${x}" = "openwrt" ]
--- a/setup.sh
+++ b/setup.sh
@@ -164,6 +164,7 @@ main() {
 	# Generate a launcher.sh file from the launcher.sh.template file
 	sed -e "s|%%SCRIPT_PREFIX%%|${SCRIPT_PREFIX}|g" -e "s|%%CONFIG_PREFIX%%|${CONFIG_PREFIX}|g" \
 		"${tmp}/launcher.sh.template" > "${SCRIPT_PREFIX}/launcher.sh"
+	chmod +x "${SCRIPT_PREFIX}/launcher.sh"
 
 	# Also for OpenWrt generate the service file from cake-autorate.template but DO NOT ACTIVATE IT
 	if [ "${MY_OS}" = "openwrt" ]
--- a/setup.sh
+++ b/setup.sh
@@ -203,7 +203,7 @@ main() {
 			;;
 		"asuswrt")
 			printf '%s\n' "Launch script on boot with:"
-			printf '%s\n\n' "   echo ${SCRIPT_PREFIX}/launcher.sh >> /jffs/scripts/post-mount"
+			printf '%s\n\n' "   echo ${SCRIPT_PREFIX}/launcher.sh >> /jffs/scripts/qos-start"
 			;;
 		*)
 			;;
--- a/INSTALLATION.md
+++ b/INSTALLATION.md
@@ -1,11 +1,11 @@
-# Installing cake-autorate on OpenWrt
+# Installing cake-autorate
 
 **cake-autorate** is a script that minimizes latency by adjusting CAKE
 bandwidth settings based on traffic load and round-trip time
 measurements. See the main [README](./README.md) page for more details
 of the algorithm.
 
-## Installation Steps
+## Installation Steps (OpenWrt)
 
 cake-autorate provides an installation script that installs all the
 required tools. To use it:
@@ -36,21 +36,50 @@ required tools. To use it:
 - The installer script will detect a previous configuration file, and
   ask whether to preserve it.
 
+## Installation Steps (Asus Merlin)
+
+- [SSH into the router](https://github.com/RMerl/asuswrt-merlin.ng/wiki/SSHD)
+
+- Make sure these are installed: entware; bash; iputils-ping; and fping.
+
+- Use the installer script by copying and pasting each of the commands
+  below. The commands retrieve the current version from this repo:
+
+  ```bash
+  wget -O /tmp/cake-autorate_setup.sh https://raw.githubusercontent.com/lynxthecat/cake-autorate/master/setup.sh
+
+  sh /tmp/cake-autorate_setup.sh
+  ```
+
+## Initial Configuration Steps (OpenWrt and Asus Merlin)
+
 - For a fresh install, you will need to undertake the following steps.
 
 - Edit the _config.primary.sh_ script (in the _/root/cake-autorate_
-  directory) using vi or nano to set the configuration parameters
-  below (see comments in _config.primary.sh_ for details).
+  directory for OpenWrt or in the _/jffs/scripts/cake-autorate_ 
+  direction for Asus Merlin) using vi or nano to set the configuration 
+  parameters below (see comments in _config.primary.sh_ for details).
 
   - Change `dl_if` and `ul_if` to match the names of the upload and
-    download interfaces to which CAKE is applied. These can be
-    obtained, for example, by consulting the configured SQM settings
-    in LuCI or by examining the output of `tc qdisc ls`.
+    download interfaces to which CAKE is applied. 
 
     | Variable | Setting                                          |
     | -------: | :----------------------------------------------- |
     |  `dl_if` | Interface that downloads data (often _ifb4-wan_) |
     |  `ul_if` | Interface that uploads (often _wan_)             |
+
+
+  - For OpenWrt installations, these can be obtained, for example, 
+    by consulting the configured SQM settings in LuCi or by examining 
+    the output of `tc qdisc ls`.
+
+  - For Asus Merlin the requisite interfaces can also be obtained
+    by examining the output of `tc qdisc ls`. These are most likely:
+  
+    ```bash
+    dl_if=ifb4eth0 # download interface
+    ul_if=eth0     # upload interface
+    ```
 
   - Choose whether cake-autorate should adjust the shaper rates
     (disable for monitoring only):
@@ -189,7 +218,7 @@ cd /root/cake-autorate     # to the cake-autorate directory
   upload rates as you use the connection.
 - Press ^C to halt the process.
 
-## Install as a service
+## Install as a service (OpenWrt)
 
 You can install cake-autorate as a service that starts up the autorate
 process whenever the router reboots. To do this:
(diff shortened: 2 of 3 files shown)


LiveWire1968:
@lynxthecat  I still need to change the shebang to bash, besides that it seems to work.

lynxthecat:
@LiveWire1968 does the launcher and suggested lines to post-mount work?

Regarding the shebang, why isn't: #!/bin/sh correct? Does Asus Merlin not include /bin/sh?

rany2:
> Regarding the shebang, why isn't: `#!/bin/sh` correct? Does Asus Merlin not include /bin/sh?

It's correct but the busybox version they ship is old and has bugs. You could see one of the bug fixes I made with the `set -e; . /etc/os-release; set +e` trickery I had to do.

LiveWire1968:
@lynxthecat  rany said its a bug with busybox sh, I thought i saw he had did a change on it, but i cant find it again.
echo /jffs/scripts/launcher.sh >> /jffs/scripts/post-mount   needs changed to 
echo /jffs/scripts/cake-autorate/launcher.sh >> /jffs/scripts/post-mount . and launcher.sh needs chmod +x

LiveWire1968:
> > Regarding the shebang, why isn't: `#!/bin/sh` correct? Does Asus Merlin not include /bin/sh?
> 
> It's correct but the busybox version they ship is old and has bugs. You could see one of the bug fixes I made with the `set -e; . /etc/os-release; set +e` trickery I had to do.

@rany2 the set -e doesnt seem to work

rany2:
@LiveWire1968 it's the opposite, should be `set +e; . /etc/os-release; set -e`

LiveWire1968:
admin@RT-AX86U-D7D0:/tmp/home/root# sh -x setup.sh
+ main
+ set -eu
+ DEPENDENCIES=jsonfilter wget tar grep bash
+ REPOSITORY=lynxthecat/cake-autorate
+ BRANCH=master
+ SRC_DIR=https://github.com/lynxthecat/cake-autorate/archive/
+ API_URL=https://api.github.com/repos/lynxthecat/cake-autorate/commits/master
+ DOC_URL=https://github.com/lynxthecat/cake-autorate/tree/master#installation-on-openwrt
+ SCRIPT_PREFIX=
+ CONFIG_PREFIX=
+ MY_OS=unknown
+ unset ID_LIKE
+ set +e
+ . /etc/os-release

rany2:
@LiveWire1968 replace the `.` with `source` maybe?

rany2:
I know but it shouldn't be a fatal error, but I presume that you tried to use `source` instead and it didn't work either

LiveWire1968:
> I know but it shouldn't be a fatal error, but I presume that you tried to use `source` instead and it didn't work either

weird bash works even with the error

rany2:
yeah it's another busybox sh bug, i guess it might be worth it to just guard that with a `[ -a ` check

rany2:
I'm curious does AsusWRT-Merlin ship with another sh? Does `/usr/bin/env sh ./setup.sh` behave differently?

LiveWire1968:
> I'm curious does AsusWRT-Merlin ship with another sh? Does `/usr/bin/env sh ./setup.sh` behave differently?

env sh is still the same busybox, so this doesnt work either, there is no other sh

rany2:
> yeah it's another busybox sh bug, i guess it might be worth it to just guard that with a `test -a ` check

@lynxthecat just do that then

lynxthecat:
You mean something like this:

```
 33         # Check if OS is OpenWRT or derivative
 34         unset ID_LIKE
 35
 36         [ -f /etc/os-release ] && . /etc/os-release
```

rany2:
@lynxthecat use `-e` instead, some OpenWRT versions have os-release with executable bit set

setup.sh
```diff
@@ -190,8 +187,20 @@ main() {
 	printf '\n%s\n\n' "${version} successfully installed, but not yet running"
 	printf '%s\n' "Start the software manually with:"
 	printf '%s\n' "   cd ${SCRIPT_PREFIX}; ./cake-autorate.sh"
-	printf '%s\n' "Run as a service with:"
-	printf '%s\n\n' "   service cake-autorate enable; service cake-autorate start"
+
+	case "$MY_OS" in
```
rany2, line 191: I'm not sure if we setup shellcheck to care, but this should be "${MY_OS}" so stay consistent with the rest of the style.

setup.sh
```diff
@@ -81,16 +85,6 @@ main() {
 	done
 	[ "${exit_now}" -ge 1 ] && exit "${exit_now}"
 
-	# Retrieve required packages if not present
```
rany2, line 84: Keep opkg as it is provided by entware

setup.sh
```diff
@@ -14,7 +14,7 @@ main() {
 	set -eu
 
 	# Setup dependencies to check for
-	DEPENDENCIES="jsonfilter wget tar grep"
+	DEPENDENCIES="jsonfilter wget tar grep bash"
```
rany2, line 17: Should be removed when adding back `opkg`

LiveWire1968:
> @LiveWire1968 could you test again and verify all is working (including the revised post-mount guidance)?

@lynxthecat this seems to work now, the echo looks correct. Launcher.sh still needs to be excutable.

setup.sh
rany2: I think it is better to check for `fping` in the DEPENDENCIES then given that it's the default option and superior to iputils-ping. IMO iputils-ping should be dropped completely from cake-autorate

rany2: I don't feel strongly either way about this

lynxthecat:
Does it work now?

rany2:
For the record the only reason why the previous PR took time is because I wanted to make a proper entware service for cake-autorate NOT this hack in post-mount

lynxthecat:
> For the record the only reason why the previous PR took time is because I wanted to make a proper entware service for cake-autorate NOT this hack in post-mount

I guess this is the Asus Merlin way though:

https://github.com/RMerl/asuswrt-merlin.ng/wiki/User-scripts

rany2:
> > For the record the only reason why the previous PR took time is because I wanted to make a proper entware service for cake-autorate NOT this hack in post-mount
> 
> I guess this is the Asus Merlin way though:
> 
> https://github.com/RMerl/asuswrt-merlin.ng/wiki/User-scripts

It's not proper though, take a look at https://github.com/RMerl/asuswrt-merlin.ng/blob/master/release/src-rt-5.02axhnd/targets/fs.src/rom/etc/init.d/mount-fs.sh. The issue with our init.d right now is that it's sourcing rc.common, we need to make it work for both or create a different ASUSWRT variant. I can't find another link atm but there is another package that does this properly too..

lynxthecat:
@RMerl any thoughts on how we are doing things here?

rany2:
IMO the URL you linked to is meant for trivial one-liner scripts, not cake-autorate :)

LiveWire1968:
> IMO the URL you linked to is meant for trivial one-liner scripts, not cake-autorate :)
@rany there are some fairly complicated asus-merlin script packages, they all get started out of one of the startup scripts
in /jffs/scripts never init.d

rany2:
> there are some fairly complicated asus-merlin script packages, they all get started out of one of the startup scripts
> in /jffs/scripts never init.d

how would you restart them/handle failure? you need some kind of process monitor at a minimum, it seems shortsighted/too simple to just fork from that file and call it a day but I guess given that this is the norm at AsusWRT-Merlin you could just do this too...

RMerl:
Why not use qos-start?  That script gets called whenever QoS is initialized.  That means you will be able to have it only configure things if QoS is starting with Cake actually enabled as the scheduler.

lynxthecat:
Could you elaborate? What does qos-start look like these days?

RMerl:
> Could you elaborate? What does `qos-start` look like these days?

It doesn't look like anything.  It's a script that a user can create to launch tasks whenever QoS gets initialized.

https://github.com/RMerl/asuswrt-merlin.ng/wiki/User-scripts#qos-start

lynxthecat:
So if I'm understanding correctly, then we can give this guidance:

```
"asuswrt")
	printf '%s\n' "Launch script on boot with:"
	printf '%s\n\n' "   echo ${SCRIPT_PREFIX}/launcher.sh >> /jffs/scripts/qos-start"
	;;
```

rany2:
@RMerl Personally, I would have preferred a regular service file as I want cake-autorate to be monitored and for cake-autorate to be restartable for when it's updated, etc. It's a bit strange that this is just not encouraged in AsusWRT-Merlin but I guess it's a simplicity vs robustness trade-off. Just forking will always be more straightforward !

rany2 (approved):

LiveWire1968:
> > Could you elaborate? What does `qos-start` look like these days?
> 
> It doesn't look like anything. It's a script that a user can create to launch tasks whenever QoS gets initialized.
> 
> https://github.com/RMerl/asuswrt-merlin.ng/wiki/User-scripts#qos-start

It needs entware, why i recommended post-mount

RMerl:
> @RMerl Personally, I would have preferred a regular service file as I want cake-autorate to be monitored and for cake-autorate to be restartable for when it's updated, etc. It's a bit strange that this is just not encouraged in AsusWRT-Merlin but I guess it's a simplicity vs robustness trade-off. Just forking will always be more straightforward !

Asuswrt is simply not designed around a SysV init system, the entire operating system is monolithic.  Having hooks like these scripts is the only way to interface with it.

rany2:
> > @RMerl Personally, I would have preferred a regular service file as I want cake-autorate to be monitored and for cake-autorate to be restartable for when it's updated, etc. It's a bit strange that this is just not encouraged in AsusWRT-Merlin but I guess it's a simplicity vs robustness trade-off. Just forking will always be more straightforward !
> 
> Asuswrt is simply not designed around a SysV init system, the entire operating system is monolithic. Having hooks like these scripts is the only way to interface with it.

Oh, I thought that I could adapt my init script to not call rc.common as a workaround to have it work on Asuswrt. I didn't know that this was just not possible. It's a shame.

lynxthecat:
@LiveWire1968 do the installation instructions for Asus Merlin look OK to you?

LiveWire1968:
@lynxthecat  I was testing this again this morning. I cannot get the latest cake-autorate.sh or launcher.sh to start
from qos-start or post-mount, cant get any log of it when booting the router. 

What is working

echo /jffs/scripts/cake-autorate/launcher.sh >/opt/etc/init.d/S99cake-autorate
chmod +x /opt/etc/init.d/S99cake-autorate   

Then everything starts up fine....

lynxthecat:
@rany2 any idea? Permissions issue?

rany2:
@lynxthecat most likely, read https://github.com/RMerl/asuswrt-merlin.ng/wiki/User-scripts#creating-scripts

lynxthecat:
So just need to add suggestion to chmod +x at:

https://github.com/lynxthecat/cake-autorate/blob/a45dbf3e4a80be8a311ae325d2acb39596c91055/setup.sh#L237

rany2:
@lynxthecat they specifically suggest `chmod a+rx` most likely because of a default restrictive umask, so he should try that not `chmod +x`

LiveWire1968:
> So just need to add suggestion to chmod +x at:
> 
> https://github.com/lynxthecat/cake-autorate/blob/a45dbf3e4a80be8a311ae325d2acb39596c91055/setup.sh#L237
> 
> @LiveWire1968 would simple chmod +x there fix it?

no, i had it set excutable, i put a test in qos-start, its working fine

LiveWire1968:
and fyi, its not working from post-mount either, which is where entware and other scripts startup.. Id guess another environment problem, but i havent figured it out

lynxthecat:
Did you try the chmod a+rx?

lynxthecat:
Could you also try adding inside qos-start or post-mount the line:

```
env > /tmp/env_output.txt 
```

And paste the output from that?

LiveWire1968:
admin@RT-AX86U-D7D0:/tmp# cat env_output.txt
router=100.64.0.1
subnet=255.192.0.0
USER=root
staticroutes=192.168.100.1/32 0.0.0.0 34.120.255.244/32 0.0.0.0 0.0.0.0/0 100.64.0.1
SHLVL=1
LD_LIBRARY_PATH=/lib:/usr/lib:/lib/aarch64
HOME=/
ENV=/etc/profile
interface=eth0
dns=1.1.1.1 8.8.8.8
siaddr=10.10.10.10
serverid=100.64.0.1
TERM=vt100
mtu=1500
PATH=/sbin:/bin:/usr/sbin:/usr/bin:/opt/sbin:/opt/bin
ip=100.74.88.2
lease=300
mask=10
SHELL=/bin/sh
PWD=/tmp/bwdpi
opt53=05
TZ=GMT5
TMOUT=0

lynxthecat:
And you definitely tried chmod a+rx? 

Any other thoughts @rany2?

lynxthecat:
@RMerl can you advise here?

LiveWire1968:
I added env > /tmp/cake-test to the top of cake-autorate.sh .. the script is trying to run
admin@RT-AX86U-D7D0:/tmp# cat cake-test
subnet=255.192.0.0
router=100.64.0.1
USER=root
staticroutes=192.168.100.1/32 0.0.0.0 34.120.255.244/32 0.0.0.0 0.0.0.0/0 100.64.0.1
LD_LIBRARY_PATH=/lib:/usr/lib:/lib/aarch64
SHLVL=2
HOME=/
ENV=/etc/profile
interface=eth0
siaddr=10.10.10.10
dns=1.1.1.1 8.8.8.8
serverid=100.64.0.1
TERM=vt100
mtu=1500
PATH=/sbin:/bin:/usr/sbin:/usr/bin:/opt/sbin:/opt/bin
ip=100.74.88.2
SHELL=/bin/sh
mask=10
lease=300
PWD=/tmp/bwdpi
opt53=05
TZ=GMT5
TMOUT=0

LiveWire1968:
Not sure this helps, when i change the line in qos-start to 
/jffs/scripts/cake-autorate/cake-autorate.sh 2> /tmp/cak2-test    I get
admin@RT-AX86U-D7D0:/tmp# cat cak2-test
/jffs/scripts/cake-autorate/cake-autorate.sh: line 35: syntax error: unexpected redirection

rany2:
@LiveWire1968 might be a bug in AsusWRT, that suggests that the wrong shell is being used and the shebang is getting ignored

LiveWire1968:
> @LiveWire1968 might be a bug in AsusWRT, that suggests that the wrong shell is being used and the shebang is getting ignored

I agree, I think that was the first error before the shebang stuff was changed. Even though its not the asus way, probably easiest to direct users to use entware's init.d... it works fine

lynxthecat:
Can you please suggest exactly what lines we should use for suggested action post install?

LiveWire1968:
I just retested this, restarted router.. works fine.
echo /jffs/scripts/cake-autorate/launcher.sh >/opt/etc/init.d/S99cake-autorate
chmod +x /opt/etc/init.d/S99cake-autorate

lynxthecat:
Since maybe this is a bug in Asus Merlin hopefully @RMerl can advise.

LiveWire1968:
> Since this seems to relate to a peculiarity or bug in Asus Merlin, hopefully @RMerl can advise.

He has moved over to openwrt, but @dave14305 was a scripting wizard on asuswrt.. Maybe he has some insight also.

rany2:
> > Since this seems to relate to a peculiarity or bug in Asus Merlin, hopefully @RMerl can advise.
> 
> He has moved over to openwrt, but @dave14305 was a scripting wizard on asuswrt.. Maybe he has some insight also.

If @RMerl abandoned asuswrt, then you really shouldn't be using it right now (I doubt that he did, seems he's still pushing out updates)

LiveWire1968:
> > > Since this seems to relate to a peculiarity or bug in Asus Merlin, hopefully @RMerl can advise.
> > 
> > 
> > He has moved over to openwrt, but @dave14305 was a scripting wizard on asuswrt.. Maybe he has some insight also.
> 
> If @RMerl abandoned asuswrt, then you really shouldn't be using it right now (I doubt that he did, seems he's still pushing out updates)

@rany2 I meant dave14305 has moved to openwrt.....

lynxthecat:
https://www.snbforums.com/threads/cake-autorate-adding-support-for-asus-merlin.89639/page-2

Looks like mktemp may not be available by default im Asus Merlin.

LiveWire1968:
> https://www.snbforums.com/threads/cake-autorate-adding-support-for-asus-merlin.89639/page-2
> 
> Looks like mktemp may not be available by default in Asus Merlin.

I saw that post, I already had it installed via opkg for some reason.. So missed the requirement

rany2:
> https://www.snbforums.com/threads/cake-autorate-adding-support-for-asus-merlin.89639/page-2
> 
> Looks like mktemp may not be available by default in Asus Merlin.
> 
> I wonder what would cause this line:
> 
> https://github.com/lynxthecat/cake-autorate/blob/4dea0575726fe85faf6065269f1343c9de42808c/cake-autorate.sh#L34
> 
> to throw:
> 
> /jffs/scripts/cake-autorate/cake-autorate.sh: line 34: syntax error: unexpected redirection

Obviously it's calling regular /bin/sh instead of bash.

rany2:
Btw the asuswrt-merlin branch no longer exists, why are you telling people on the forum to try to use it? Of course there won't be any output :)

LiveWire1968:
> > https://www.snbforums.com/threads/cake-autorate-adding-support-for-asus-merlin.89639/page-2
> > Looks like mktemp may not be available by default in Asus Merlin.
> > I wonder what would cause this line:
> > https://github.com/lynxthecat/cake-autorate/blob/4dea0575726fe85faf6065269f1343c9de42808c/cake-autorate.sh#L34
> > 
> > to throw:
> > /jffs/scripts/cake-autorate/cake-autorate.sh: line 34: syntax error: unexpected redirection
> 
> Obviously it's calling regular /bin/sh instead of bash.

Agreed @rany2 so I think hes using the wrong version after all your fixes

lynxthecat:
Just an old post @rany2. I’ve linked to the new installation instructions. 

I’ll add the missing dependency when I get a chance.

LiveWire1968:
> > @rany2 I meant dave14305 has moved to openwrt.....
> 
> I switch back and forth based on what interests me, usually when there’s an compelling update. Anything in particular you guys need advice on? I’m running a Merlin router this week. I was on vacation last week…

@dave14305 Im sure @rany2 will pop in. Been a issue with cake-autorate needing bash on asuswrt, been some asuswrt sh bugs found. So it runs fine now, EXCEPT, it will not start from like qos-start,etc.. the sh bugs pop back up, like startup is ignoring the shebang. it will start fine from entware init.d...

setup.sh
dave14305: > Not sure I agree. I don't think we need to / should install packages using opkg if they're missing. It's enough to report dependencies missing, no? Then user can decide to install. By the way, fping isn't strictly needed since user may instead use 'tsping' or 'iputils-ping'.

For Merlin users, you will (eventually) want to install all dependencies so there is a clean, automated installation through amtm one day. Otherwise, the friction to install will be too much for the less technical users of Merlin firmware.

Issue #293, opened 2024-04-12
More congenial setup.sh updates


lynxthecat:
Could the setup script check that it is the latest version, and if not then perhaps offer to overwrite itself with the latest version? 

To help avoid situations like this:

```
BusyBox v1.36.1 (2024-04-03 00:56:56 UTC) built-in shell (ash)

  _______                     ________        __
 |       |.-----.-----.-----.|  |  |  |.----.|  |_
 |   -   ||  _  |  -__|     ||  |  |  ||   _||   _|
 |_______||   __|_____|__|__||________||__|  |____|
          |__| W I R E L E S S   F R E E D O M
 -----------------------------------------------------
 OpenWrt 23.05.3, r23809-234f1a2efa
 -----------------------------------------------------
root@OpenWrt-1:~# cd cake-autorate/
root@OpenWrt-1:~/cake-autorate# ./setup.sh
At least one instance of cake-autorate appears to be running - exiting
If you want to install a new version, first stop any running instance of cake-autorate
If you are sure that no instance of cake-autorate is running, delete the /var/run/cake-autorate directory
root@OpenWrt-1:~/cake-autorate# service cake-autorate stop
root@OpenWrt-1:~/cake-autorate# ./setup.sh
Installing cake-autorate in /root/cake-autorate...
Previous configuration present - keep it? [Y/n]
mv: can't rename '/tmp/tmp.NJfmff/launcher.sh': No such file or directory
root@OpenWrt-1:~/cake-autorate# ./setup.sh
Installing cake-autorate in /root/cake-autorate...
Previous configuration present - keep it? [Y/n]
mv: can't rename '/tmp/tmp.eihdno/launcher.sh': No such file or directory
root@OpenWrt-1:~/cake-autorate# wget https://raw.githubusercontent.com/lynxthecat/cake-autorate/master/setup.sh
Downloading 'https://raw.githubusercontent.com/lynxthecat/cake-autorate/master/setup.sh'
Connecting to 185.199.111.133:443
Writing to 'setup.sh'
setup.sh             100% |*******************************|  6806   0:00:00 ETA
Download completed (6806 bytes)
root@OpenWrt-1:~/cake-autorate# ./setup.sh
-ash: ./setup.sh: Permission denied
root@OpenWrt-1:~/cake-autorate# chmod +x setup.sh
root@OpenWrt-1:~/cake-autorate# ./setup.sh
Detected Operating System: openwrt
Installation directories for detected Operating System:
[...]

PR #301, opened 2024-05-03
Check local setup.sh script matches remote

lynxthecat:
Addresses #293

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -123,6 +123,16 @@ main() {
 	wget -qO- "${SRC_DIR}/${commit}.tar.gz" | tar -xozf - -C "${tmp}"
 	mv "${tmp}/cake-autorate-"*/* "${tmp}"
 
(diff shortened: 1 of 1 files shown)


lynxthecat:
How about this @rany2?

rany2:
Looks good but the usage of cat here is useless. You could just pass the filename as argument to sha256sum.

rany2:
Also I'm not sure if ASUSWrt has that.

rany2 (changes requested):
Pass as argument to sha256sum and add to dependencies

rany2:
My plan was to just have it be self replacing but this works ok too

rany2:
Also errors should be printed to stderr.

rany2:
I checked and you could just use `cmp` instead of `sha256sum`.

rany2:
Just close this I'll do it myself.

Issue #244, opened 2023-07-13
Have a maint script that automatically creates the "Status" text


rany2:

lynxthecat:
The "Status" text?

rany2:
https://github.com/lynxthecat/cake-autorate#status

lynxthecat:
How about this? My sed skills admittedly leave a little to be desired.

lynxthecat:
Does this work:

https://github.com/lynxthecat/cake-autorate/blob/eebf9abe3f706e97e24830fe083f00d8739ce12b/maint/update_setupsh_branch.sh#L19

Is the idea to checkout new branch v3.1, then run:

```
 ./maint/update_setupsh_branch.sh "v3.2"
```

because the sed for me doesn't seem to work.w 

But perhaps I'm missing something.

PR #304, opened 2024-05-13
Update version branch link in README.md

lynxthecat:
See #244


rany2:
It looks like it would work but I haven't tested it

rany2 (approved):
Approved assuming this works as expected.

PR #314, opened 2024-08-12
Monitor and log CPU usage

lynxthecat:
Leverage /proc/stat to support monitoring and logging CPU use across detected cores.


lynxthecat:
@rany2 I wonder if you see any issues or improvements to my code here?

cake-autorate.sh
```diff
@@ -1495,6 +1533,42 @@ do
 				set_shaper_rate "dl"
 				set_shaper_rate "ul"
 
+				# update CPU load if CPU monitoring interval exceeded
+				if (( output_cpu_stats || output_cpu_raw_stats ))
+				then
+					if (( t_start_us > t_last_cpu_usage_check_us + monitor_cpu_usage_interval_us ))
```
rany2, line 1539: unnecessary nesting, could be combined into one if statement

rany2 (changes requested):

lynxthecat:
How about now?

rany2 (approved):
Not sure how useful these metrics would be, but we could easily undo it if it causes problems.

PR #336, opened 2025-07-22
Add support for irtt

lrussell887:
Adds support for `irtt` as a means of getting reliable OWD data.

A package for `irtt` is not yet present in OpenWrt stable but is available in snapshots. For now, it must be statically compiled and the binary placed in `/usr/bin`.

Due to `irtt`'s design as a test-runner rather than a daemon, we parse its line output rather than generating JSON. We also need to periodically restart it to prevent excessive memory usage since all test results are kept in-memory. This is done through the new `irtt_session_duration_m` variable which controls the test duration. After the duration elapses, irtt is immediately re-spawned. 

The use of `stdbuf -oL` is to enforce line-buffering when run non-interactively, as opposed to block-buffering which could otherwise delay processing until the stdio buffer is filled. This adds a dependency on `coreutils-stdbuf` in OpenWrt.

Checking if `start_time` was less than three seconds ago prevents spinning due to `irtt` exiting prematurely for any reason, such as an internet outage or reflector issue.

The `_convert_to_us` function rejects erroneous data, such as negative numbers or very large values (e.g. `-4835h58m3s`, `4835h58m3s`) caused by clock skew. It checks the latter by assuming if the string contains `h` or `m` (after checking `ms`) it must be invalid. This is also how invalid data lines are rejected, as they would fail to be parsed.

I don't believe the timestamp rollover issue compensated for in `tsping` is relevant to `irtt` since it uses epoch time, so it was not copied over.

I attempted to adhere the style and structure of existing pingers as closely as possible while working around `irtt`'s limitations, Please let me know if any changes need to be made.

### DIFF
--- a/INSTALLATION.md
+++ b/INSTALLATION.md
@@ -300,11 +300,14 @@ adjusting logging parameters such as `log_file_max_time_mins` or
 ## Launch on Boot (Asus Merlin)
 
 cake-autorate can be launched on boot by adding an appropriate entry
-to e.g. post-mount or qos-start - see
+to e.g. post-mount - see
 [here](https://github.com/RMerl/asuswrt-merlin.ng/wiki/User-scripts).
 
+For example, add these lines to /jffs/scripts/post-mount:
+
 ```bash
-echo /jffs/scripts/cake-autorate/launcher.sh >> /jffs/scripts/qos-start
+source /etc/profile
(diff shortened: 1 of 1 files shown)


lrussell887:
Accidentally broke git history merging commits, opened new request #337 .

PR #337, opened 2025-07-22
Add support for irtt

lrussell887:
Adds support for `irtt` as a means of getting reliable OWD data.

A package for `irtt` is not yet present in OpenWrt stable but is available in snapshots. For now, it must be statically compiled and the binary placed in `/usr/bin`.

Due to `irtt`'s design as a test-runner rather than a daemon, we parse its line output rather than generating JSON. We also need to periodically restart it to prevent excessive memory usage since all test results are kept in-memory. This is done through the new `irtt_session_duration_m` variable which controls the test duration. After the duration elapses, irtt is immediately re-spawned. 

The use of `stdbuf -oL` is to enforce line-buffering when run non-interactively, as opposed to block-buffering which could otherwise delay processing until the stdio buffer is filled. This adds a dependency on `coreutils-stdbuf` in OpenWrt.

Checking if `start_time` was less than three seconds ago prevents spinning due to `irtt` exiting prematurely for any reason, such as an internet outage or reflector issue.

The `_convert_to_us` function rejects erroneous data, such as negative numbers or very large values (e.g. `-4835h58m3s`, `4835h58m3s`) caused by clock skew. It checks the latter by assuming if the string contains `h` or `m` (after checking `ms`) it must be invalid. This is also how invalid data lines are rejected, as they would fail to be parsed.

I don't believe the timestamp rollover issue compensated for in `tsping` is relevant to `irtt` since it uses epoch time, so it was not copied over.

I attempted to adhere the style and structure of existing pingers as closely as possible while working around `irtt`'s limitations, Please let me know if any changes need to be made.


lrussell887:
The function to convert time strings to microseconds (`time_str_to_us`) has been moved to `lib.sh` so it is not declared on every iteration.

cake-autorate.sh
```diff
@@ -494,6 +494,22 @@ start_pinger()
 
 	case ${pinger_binary} in
 
+		irtt)
+			run_irtt_pinger_instance() {
+				while :; do
+					local start_time=${EPOCHSECONDS}
+					while IFS= read -r line; do
```
lynxthecat, line 501: This will be costly because bash needs to read byte by byte to identify each end of line character. It's something I've wrestled a lot with in cake-autorate. 

See here:

https://www.reddit.com/r/bash/comments/1bilrb2/comment/l2yz3e8/?context=3

And here:

https://github.com/lynxthecat/cake-autorate/commit/1304b19514d7c7bb3241065c1a943f3f6b084f2f

Did you consider having awk handle the processing to add the requisite initial columns?

lrussell887: The utility of `run_irtt_pinger_instance()` is that it handles the respawning of `irtt` after its test duration elapses. This makes `irtt` appear as a continuous data source to the main loop, preventing the health-check from (correctly) flagging it as a misbehaving reflector and triggering a rotation. I don't think we can get rid of the wrapper entirely due to the nature of how `irtt` works.

Exploring your suggestion to use an `awk` pipeline to avoid the `while read` loop reveals an issue. While we can solve for `irtt` not providing the reflector IP by passing it in with the `-v` flag, the problem is time-stamping. We are simply unable to get a timestamp with at least millisecond precision directly from `awk` or `gawk` based on my testing. Having `awk` make an external call to `date +%s%N` (which would require installing `coreutils-date`) would likely be more expensive than the read loop itself due to process-spawning overhead for every single line processed. This leaves `bash`'s `${EPOCHREALTIME}` as the only sensible source for a timestamp of the required precision.

I also considered time-stamping later in the main processing loop as you suggested, but I’m not so sure that’s a good idea. This would create unpredictable "jitter" from the script's own processing, since it runs single-threaded. Instead of measuring "when the ping was received by `irtt`", we’d be measuring "when the main loop gets around to processing the ping response". Do you think this would be a problem in practice?

This seems to leave the `while read` loop inside the wrapper as the most direct approach, as it's the only way I can see to combine the `irtt` data with both the reflector IP and a reliable timestamp as close to the source as possible. Unless there is something I’m missing?

lrussell887: I've been doing some thinking and the solution to this might be to use a background shell process as a timestamp provider, and have `awk` read/write to the pipes in order to get a timestamp. Doing this allows us to avoid `bash`'s byte-by-byte reads and allows `awk` to retrieve a high-precision timestamp without spawning a new process each time. Doing this creates a dependency on `gawk` however because of `fflush()`.

Here's a basic demo:
```bash
#!/bin/bash
REQ_PIPE="/tmp/req.$$"
RESP_PIPE="/tmp/resp.$$"
mkfifo "$REQ_PIPE" "$RESP_PIPE"

{
  while read -r; do
    printf '%s\n' "${EPOCHREALTIME/.}"
  done <"$REQ_PIPE" >"$RESP_PIPE"
} &

awk -v req="$REQ_PIPE" -v resp="$RESP_PIPE" '
  BEGIN {
    for (i = 0; i < 20; i++) {
      print "" > req
      fflush(req)
      getline ts < resp
      print ts
    }
  }'

rm -f "$REQ_PIPE" "$RESP_PIPE"
```

Do you have any objections to this method? It certainly adds complexity and would probably necessitate having a `irtt_parser.awk` file just to keep things clean, but the performance gain from avoiding the second `while read` loop may justify it. In fact, since we'd already be in `awk`, we could also have it handle all the `irtt` data parsing, which would let us remove the `time_str_to_us()` function from `lib.sh`.

lrussell887: I've constructed a benchmark script to see how these two methods compare in practice.

```bash
#!/bin/bash
COUNT=100000

echo "--- METHOD A: bash 'while read' loop ---"
time {
  yes "test data from irtt" | head -n $COUNT | while IFS= read -r line; do
    printf '%s %s %s\n' "${EPOCHREALTIME/.}" "1.1.1.1" "$line"
  done > /dev/null
}

echo ""
echo "--- METHOD B: awk with bash timestamp server ---"

REQ_PIPE="/tmp/req.$$"
RESP_PIPE="/tmp/resp.$$"
mkfifo "$REQ_PIPE" "$RESP_PIPE"

{
  while read -r; do
    echo "${EPOCHREALTIME/.}"
  done <"$REQ_PIPE" >"$RESP_PIPE"
} &

time {
  yes "test data from irtt" | head -n $COUNT | gawk -v req="$REQ_PIPE" -v resp="$RESP_PIPE" -v reflector="1.1.1.1" '
    {
      print "" > req
      fflush(req)
      getline ts < resp
      print ts, reflector, $0
    }' > /dev/null
}

rm -f "$REQ_PIPE" "$RESP_PIPE"
```

These results are from my main router (with an AMD RX-427BB):
```bash
root@router:~# # 100 loops
root@router:~# ./test.sh
--- METHOD A: bash 'while read' loop ---

real    0m0.013s
user    0m0.003s
sys     0m0.012s

--- METHOD B: awk with bash timestamp server ---

real    0m0.012s
user    0m0.016s
sys     0m0.001s
root@router:~# # 1000 loops
root@router:~# ./test.sh
--- METHOD A: bash 'while read' loop ---

real    0m0.086s
user    0m0.071s
sys     0m0.019s

--- METHOD B: awk with bash timestamp server ---

real    0m0.058s
user    0m0.019s
sys     0m0.007s
root@router:~# # 10000 loops
root@router:~# ./test.sh
--- METHOD A: bash 'while read' loop ---

real    0m0.910s
user    0m0.691s
sys     0m0.321s

--- METHOD B: awk with bash timestamp server ---

real    0m0.516s
user    0m0.602s
sys     0m0.089s
root@router:~# # 100000 loops
root@router:~# ./test.sh
--- METHOD A: bash 'while read' loop ---

real    0m9.436s
user    0m7.044s
sys     0m3.675s

--- METHOD B: awk with bash timestamp server ---

real    0m5.081s
user    0m5.528s
sys     0m1.278s
```

And these results from a device with, well.. one of the worst possible CPUs still technically capable of running modern Linux:
```bash
root@wm8505:~# cat /proc/cpuinfo
processor       : 0
model name      : ARM926EJ-S rev 5 (v5l)
BogoMIPS        : 174.48
Features        : swp half thumb fastmult edsp java
CPU implementer : 0x41
CPU architecture: 5TEJ
CPU variant     : 0x0
CPU part        : 0x926
CPU revision    : 5

Hardware        : VIA/Wondermedia SoC (Device Tree Support)
Revision        : 0000
Serial          : 0000000000000000
root@wm8505:~# uname -a
Linux wm8505 6.1.144-wm8505+ #1 Sat Jul 12 20:11:01 EDT 2025 armv5tejl GNU/Linux
```
```bash
root@wm8505:~# # 100 loops
root@wm8505:~# ./test.sh
--- METHOD A: bash 'while read' loop ---

real    0m0.332s
user    0m0.220s
sys     0m0.120s

--- METHOD B: awk with bash timestamp server ---

real    0m0.428s
user    0m0.090s
sys     0m0.190s
root@wm8505:~# # 1000 loops
root@wm8505:~# ./test.sh
--- METHOD A: bash 'while read' loop ---

real    0m2.542s
user    0m2.060s
sys     0m0.480s

--- METHOD B: awk with bash timestamp server ---

real    0m2.681s
user    0m0.530s
sys     0m0.440s
root@wm8505:~# # 10000 loops
root@wm8505:~# ./test.sh
--- METHOD A: bash 'while read' loop ---

real    0m45.474s
user    0m17.920s
sys     0m27.530s

--- METHOD B: awk with bash timestamp server ---

real    0m25.188s
user    0m19.570s
sys     0m5.610s
```

So it looks like once the startup cost is paid, the IPC overhead is a lesser cost overall. Should I go ahead and implement the `irtt` pinger this way?

lynxthecat: This seems super intriguing. I'm struggling to understand your demo:

```
#!/bin/bash
REQ_PIPE="/tmp/req.$$"
RESP_PIPE="/tmp/resp.$$"
mkfifo "$REQ_PIPE" "$RESP_PIPE"

{
  while read -r; do
    printf '%s\n' "${EPOCHREALTIME/.}"
  done <"$REQ_PIPE" >"$RESP_PIPE"
} &

awk -v req="$REQ_PIPE" -v resp="$RESP_PIPE" '
  BEGIN {
    for (i = 0; i < 20; i++) {
      print "" > req
      fflush(req)
      getline ts < resp
      print ts
    }
  }'

rm -f "$REQ_PIPE" "$RESP_PIPE"
```

Surely the `while read -r` still results in byte-by-byte reads as bash's read implementation has to check each byte read to determine whether it is the newline character? I must be missing something here.

lynxthecat: Took me some time to figure out but this is cute!

lrussell887: > Did you consider gnu awk:
> 
> ```
> gawk -e '@load "time"; BEGIN { for (i = 0; i < 20; i++) { ts = gettimeofday(); printf "%d\n", ts

Oh... yes I am using gnu awk, or `gawk`. It was a requirement to have access to `fflush()`. I tried testing earlier and couldn't find a way to get an accurate enough timestamp natively. I saw the time extension but I must've been misusing it, leading me to the other solution. I suppose there's no need for the additional complexity of passing a timestamp in from bash now since that works.

lynxthecat: BTW I can't take credit for gettimeofday() - I used Grok to help me figure out your demo script concept and it suggested the use of that function.

rany2 (changes requested):
This looks good but I have yet to test it. I'm a bit unsure about how useful `irtt` is to most users as it appears that you need to spin up an `irtt` server to make use it and no public servers exist.

rany2:
Well, it doesn't seem too bad:

```
$ go build -ldflags='-s -w' ./cmd/irtt/
$ ls -lh irtt
-rwxrwxr-x 1 debian debian 3.2M Jul 23 15:32 irtt
```

cake-autorate.sh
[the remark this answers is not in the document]
lrussell887: Apologies, this has been resolved.

lrussell887:
I have updated the script to use `gawk` as an intermediary parser between `irtt` and the main script. The bash `while read` loop is no longer present as a consequence, improving efficiency. This however means now both `coreutils-stdbuf` and `gawk` are dependencies.

No orphaned child processes are left behind through the use of `SIGPIPE`. The PID of `gawk` is captured as the pinger PID. When killed, the pipe is broken, so the next time `irtt` tries to write to it it receives `SIGPIPE` and exits. The loop explicitly checks for this exit code (128 + 13) and breaks, making sure that the bash process is not left behind. If `irtt` exits for any other reason, it is re-spawned with a start time check to avoid spinning. This allows its behavior to match that of other pingers which don't exit prematurely if no ping responses are received. This is the best solution I could come up with that doesn't rely on traps or modifying the kill/terminate functions to handle process groups.

The `gawk` script is passed in using heredoc on file descriptor 3. This is safe because each `gawk` instance inherits a copy of the parent's file descriptor table, meaning each `gawk` process now has its own fd 3 pointing to its own script. This avoids creating a separate `irtt_parser.awk` file which would then need to be included in the `setup.sh` script. Doing it in-line like this felt cleaner to me, but I'm open to opinions.

lrussell887:
> Have you checked that this works reliably whilst cake-autorate is running as a service? Because the pipe behaviour didn’t work as far as I remember then as processes are handled rather differently in the service context.

Yes, I have it running as a service on my router. The `gawk`, `irtt`, and `bash` processes reliably stop as they should when the connection goes idle and when the service is stopped.

I think the behavior described in that post might be due to block buffering when run as a service. The process writing the pipe has to actually flush its buffers in order to receive `SIGPIPE`. If `ping` in that example never populates it with enough data to do so, it will never get `SIGPIPE` and exit. I enforce line buffering on both `irtt` and `gawk` using `stdbuf -oL` and `fflush("")` respectively. Meaning `irtt` is guaranteed to receive the signal on the very next line it tries to print after `gawk` dies.

cake-autorate.sh
```diff
@@ -494,6 +494,48 @@ start_pinger()
 
 	case ${pinger_binary} in
 
+		irtt)
+			# Supervises irtt as a daemon, using a kernel SIGPIPE (141) as the implicit stop signal
+			supervise_irtt_client() {
```
lynxthecat, line 499:
```suggestion
			supervise_irtt_client() 
			{
```

cake-autorate.sh
```diff
@@ -494,6 +494,48 @@ start_pinger()
...
+				while :; do
+					local start_time=${EPOCHSECONDS}
+					${ping_prefix_string} stdbuf -oL irtt client -i "${reflector_ping_interval_s}s" -d "${irtt_session_duration_m}m" ${ping_extra_args} "${reflectors[$1]}" 2>/dev/null
+					[[ $? -eq 141 ]] && break
+					(( EPOCHSECONDS - start_time < 3 )) && sleep_s 1
+				done
```
lynxthecat, line 505: Why '3' specifically?

cake-autorate.sh
```diff
@@ -494,6 +494,48 @@ start_pinger()
...
+				
+				@load "time"
+				
+				BEGIN { FS = "[= ]+" }
+				
+				/seq=/ && /rd=/ && /sd=/ {
+					ts = gensub(/\./, "", 1, sprintf("%.6f", gettimeofday()))
+					seq = $2; dl_owd = to_us($6); ul_owd = to_us($8)
+				
+					if (dl_owd >= 0 && ul_owd >= 0) {
```
lynxthecat, line 529:
```suggestion
					if (dl_owd >= 0 && ul_owd >= 0)
					{
```

cake-autorate.sh
```diff
@@ -494,6 +494,48 @@ start_pinger()
...
+					${ping_prefix_string} stdbuf -oL irtt client -i "${reflector_ping_interval_s}s" -d "${irtt_session_duration_m}m" ${ping_extra_args} "${reflectors[$1]}" 2>/dev/null
+					[[ $? -eq 141 ]] && break
+					(( EPOCHSECONDS - start_time < 3 )) && sleep_s 1
+				done
+			}
+		
+			sleep_until_next_pinger_time_slot "${pinger}"
+			
+			supervise_irtt_client "${pinger}" | gawk -v reflector="${reflectors[$pinger]}" -f /dev/fd/3 >&"${main_fd}" 3<<-'EOT' &
+				function to_us(val, mult) {
```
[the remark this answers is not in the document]
lrussell887: That change won't work for `awk`, it introduces a syntax error. It's a strict syntactic rule that the opening `{` must be on the same line as the control statement. `awk` treats the newline as the end of the statement.

It's technically possible to do
```awk
function to_us(val, mult) \
{
```
but it's not idiomatic. That being said, I can go with either.

cake-autorate.sh
lrussell887: It's somewhat arbitrary. All we're saying is "If the start time was less than 3 seconds ago, sleep for one second." It stops `irtt` from immediately re-spawning if it failed within 2 seconds of startup to prevent "spinning" where the `while` loop would consume the entire CPU. The advantage of having this statement is that if `irtt` exits normally after its configured duration, it restarts immediately minimizing the gap in data. Given that the minimum configurable session duration is 1 minute with how the variable is setup, it could just as easily be `5` or `45`. I just opted against `2` because that means it failed within 1 second which might be too short of a delay when accounting for startup depending on the device.

cake-autorate.sh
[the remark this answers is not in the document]
lynxthecat: Ah in that case let’s leave that instance.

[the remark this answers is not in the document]
lrussell887: I actually made a mistake here. You can do that fine with control statements like `if`, `while`. Just not for pattern blocks like `/seq=/ {` or `BEGIN {`. I'll be sure to use line breaks in the next commit. My apologies.

rany2:
> Does it work in the service procd context?
> 
> @rany2 see any issues with this? I have a feeling we considered job control at one time.

It seems OK to me, it's doing this in a subshell. I'm not a huge fan of this PR because its too complex. The workarounds to get `irtt` to play nice is a bit much IMO. It's not designed to be scriptable and have its output be machine readable.

Despite this, I'm not opposed to this change. I might do some refactoring work to make it easier/less-messy to support new pingers in the future but it might be unnecessary for now.

rany2 (approved):
> Complexity aside, I still think irtt lowers the barrier to entry for getting OWD data when using an ISP like T-Mobile which blocks ICMP Type 13. [...] The use of irtt also allows the native use of IPv6, which can be advantageous on a network like T-Mobile by avoiding 464XLAT, and its UDP packets are arguably more opaque and less likely to be managed compared to ICMP.

Sounds very reasonable. It's a shame that `irtt` isn't meant to be machine readable. Another thing that is a bit annoying about `irtt` is that it's written in Go, which isn't a good fit for routers (in some cases you might only have a couple megabytes of free space, dedicating most of it to `irtt` is sad).

At any rate, it seems there isn't a good alternative to `irtt`. Hopefully better alternatives come out or someone forks it to add a machine readable format.

PR #342, opened 2025-08-06
Minimize irtt restart delay

lrussell887:
We should stop parsing output from `irtt` and immediately respawn it once it reaches its waiting for final packets phase. This prevents us from waiting for statistics generation which is irrelevant to the use-case of `cake-autorate`, reducing the gap in data between restarts when the duration elapses.

During this 30 second test run, `irtt` waited 267.9ms before generating its final statistics. For a pinger running every 50ms, that's a long time to go without providing data.

```bash
seq=595 rtt=47.26ms rd=33.72ms sd=13.54ms ipdv=438µs
[x.x.x.x:2112] [WaitForPackets] waiting 267.9ms for final packets
seq=596 rtt=47.56ms rd=33.64ms sd=13.92ms ipdv=305µs
```
This change adds a check for the `[WaitForPackets]` message. As soon as we see it, we break the inner loop and restart `irtt`.

The loss of a single data point seems inconsequential compared to eliminating the current wait time, which by default is 3x the max RTT and can be up to 4 seconds if there's no response.

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -535,6 +535,10 @@ start_pinger()
 									fflush("")
 								}
 							}
+							else if (/WaitForPackets/)
+							{
+								break
+							}
 						}
 
 						close(irtt_cmd)


lynxthecat:
Can you elaborate on this as I understand the words but I think I'm missing the complete picture here.

lrussell887:
`irtt` waits for all of the packets to arrive before aggregating statistics at the end of its test duration. To do this, it waits by default 3x the maximum round trip time, or 4 seconds if no packets ever arrive in the case of a lost packet. 

From `irtt`'s client documentation:

> `--wait=wait`
> 
> Wait time at end of test for unreceived replies (default `3x4s`). Possible values:
> 
> | Format       | Meaning                                    |
> | :----------- | :----------------------------------------- |
> | `#xduration` | `#` times max RTT, or `duration` if no response |
> | `#rduration` | `#` times RTT, or `duration` if no response     |
> | `duration`   | A fixed duration                         |

In the current implementation, `irtt` is not restarted until the process closes on its own (after statistics generation). With this check, we stop early once it reaches the `WaitForPackets` stage and simply restart it then rather than allowing it to wait for any straggler packets to minimize the gap in data `cake-autorate` receives between session restarts since we know at this point no more useful data will be generated.

```bash
seq=595 rtt=47.26ms rd=33.72ms sd=13.54ms ipdv=438µs
[x.x.x.x:2112] [WaitForPackets] waiting 267.9ms for final packets
seq=596 rtt=47.56ms rd=33.64ms sd=13.92ms ipdv=305µs

                         Min     Mean   Median      Max  Stddev
                         ---     ----   ------      ---  ------
                RTT  39.53ms  47.93ms  47.49ms  89.29ms  6.18ms
         send delay   6.03ms  11.51ms   12.7ms  53.25ms  3.94ms
      receive delay  32.86ms  36.42ms   35.4ms   76.1ms  4.96ms
                                                               
      IPDV (jitter)   2.14µs    4.4ms   2.19ms  45.63ms  6.35ms
          send IPDV   3.42µs   2.56ms    599µs  35.45ms     4ms
       receive IPDV      9ns   2.87ms    979µs  40.99ms  5.71ms
                                                               
     send call time   46.2µs   86.4µs             122µs  10.5µs
        timer error   1.71µs    265µs             914µs   177µs
  server proc. time   6.06µs   8.48µs             128µs  7.06µs

                duration: 30.2s (wait 267.9ms)
   packets sent/received: 597/597 (0.00% loss)
 server packets received: 597/597 (0.00%/0.00% loss up/down)
     bytes sent/received: 35820/35820
       send/receive rate: 9.6 Kbps / 9.6 Kbps
           packet length: 60 bytes
             timer stats: 2/599 (0.33%) missed, 0.53% error
```

Basically, we're discarding the end section seen above by not waiting for it to generate nor do we try to parse it now.

lynxthecat:
Oh wait, I didn't consider the implementation sufficiently - at the moment it's already possible to specify 2x irtt instances in the same way that multiple pingers can be specified?

lrussell887:
Yes you can use multiple `irtt` pinger instances to multiple reflectors, where one `gawk` and `irtt` process runs per reflector. I'm using two right now for my setup. I'm trying to minimize the delay between session restarts when the duration elapses causing a momentary gap from one pinger. If you have 2+ pingers, `cake-autorate` still gets data from _some_ pinger just at a reduced rate while the other restarts. I'm using `sleep_until_next_pinger_time_slot` just the same as `ping`.

Issue #347, opened 2025-10-23
Add support for SOCKS proxy in the installation script


sc16me:
Please add support for SOCKS proxy in the installation script. In my region, GitHub is blocked on mobile internet. I have to modify the installation script from:
`wget -qO-`
to
`curl --socks5-hostname 127.0.0.1:1080 -sSL`
It would be great to be able to run the installation script with an additional parameter, for example in my case.
Thank you for your work!

lynxthecat:
I see the logic there but curl is not included as a default package on OpenWrt installations.

PR #348, opened 2025-10-23
Prefer curl over wget (uclient-fetch) in setup.sh

rany2:
This is so that users can specify proxy settings via environment variables like http_proxy, https_proxy, etc.

Fixes: https://github.com/lynxthecat/cake-autorate/issues/347


lynxthecat:
image

Issue #352, opened 2026-02-02
Add support for Home Assistant


lynxthecat:
I'm keen to add support for Home Assistant. 

<img width="1144" height="1141" alt="Image" src="https://github.com/user-attachments/assets/b9c9a24a-4dd0-49bf-9bd3-eb1a691e8103" />

I've got something that almost works:

```
#!/bin/sh /etc/rc.common

USE_PROCD=1
START=98
STOP=10

LOG_FILE="/var/log/cake-autorate.primary.log"

MQTT_HOST="192.168.x.x"
MQTT_PORT="1883"
MQTT_USER=""
MQTT_PASS=""
MQTT_TOPIC="cake-autorate"

DISC_PREFIX="homeassistant"
DEVICE_ID="openwrt"
DEVICE_NAME="OpenWrt"

MIN_INTERVAL_S=5

publish_sensor () {
mosquitto_pub -h "$MQTT_HOST" -p "$MQTT_PORT" -u "$MQTT_USER" -P "$MQTT_PASS" -r -q 1 \
-t "$DISC_PREFIX/sensor/$DEVICE_ID/$1/config" -m "$2"
}

publish_discovery() {

publish_sensor "dl_achieved_rate_kbps" \
'{"name":"CAKE DL Achieved Rate","state_topic":"cake-autorate","value_template":"{{ value_json.dl_achieved_rate_kbps }}","unit_of_measurement":"kbps","state_class":"measurement","unique_id":"openwrt_dl_achieved_rate","device":{"identifiers":["openwrt"],"name":"OpenWrt"}}'

publish_sensor "ul_achieved_rate_kbps" \
'{"name":"CAKE UL Achieved Rate","state_topic":"cake-autorate","value_template":"{{ value_json.ul_achieved_rate_kbps }}","unit_of_measurement":"kbps","state_class":"measurement","unique_id":"openwrt_ul_achieved_rate","device":{"identifiers":["openwrt"],"name":"OpenWrt"}}'

publish_sensor "cake_dl_rate_kbps" \
'{"name":"CAKE DL Rate","state_topic":"cake-autorate","value_template":"{{ value_json.cake_dl_rate_kbps }}","unit_of_measurement":"kbps","state_class":"measurement","unique_id":"openwrt_cake_dl_rate","device":{"identifiers":["openwrt"],"name":"OpenWrt"}}'

publish_sensor "cake_ul_rate_kbps" \
[...]

PR #358, opened 2026-04-09
mqtt-publisher: average samples over emit window

bernhardberger:
## Problem

The 1s throttle from #355 grabs whatever sample happens to land at emit time. For throughput data this means rates look mostly near-zero with occasional spikes when the emit catches a burst. Basically aliasing. See [this comment on #352](https://github.com/lynxthecat/cake-autorate/issues/352#issuecomment-4217379534) for context.

## What this does

Accumulates all samples over the `MIN_INTERVAL_S` window and emits the mean for numeric fields:

- **Rates/throughput** (kbps), **delays/OWD** (us), **CPU** (%): averaged over window
- CPU lines only come in at ~1Hz vs ~20Hz for SUMMARY, so when no CPU sample arrives in a window the previous window's mean carries forward
- **Load condition** and **shaper rate**: latest value wins (categorical/discrete, averaging makes no sense here)

Adds a `"samples"` field to the JSON for debugging.

Also fixes the merge conflict markers left over from #356.

Ref: #352, #355

### DIFF
--- a/mqtt-publisher.sh
+++ b/mqtt-publisher.sh
@@ -99,61 +99,71 @@ publish_stats()
         awk -F'; ' -v min_int="$MIN_INTERVAL_S" '
         BEGIN {
             last_emit = 0
-            dl_achieved_rate_kbps = ul_achieved_rate_kbps = 0
-            dl_sum_delays = ul_sum_delays = 0
-            dl_avg_owd_delta_us = ul_avg_owd_delta_us = 0
-            dl_load_condition = ul_load_condition = "unknown"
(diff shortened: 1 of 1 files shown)


lynxthecat:
@bernhardberger I also think that it would be nice to switch from seconds to milliseconds or even microseconds. What do you think?

bernhardberger:
For typical use (1-5s) seconds feels natural and readable. Max log rate is ~20Hz so anything below 50ms is pointless anyway. Milliseconds I could see, but microseconds seems like too much.

PR #361, opened 2026-04-24
Add experimental upload-only mode

Freundschaft:
please note that this pull request has been created with the assistance of chatgpt codex, and reviewed by me
the implementation is quite basic and still needs testing


lynxthecat:
Argh this is very hacky. You could use it yourself but I really don't think I would want to merge this in. AI-generated code seems to have a way of becoming very ugly fast.

Freundschaft:
Sure, ill use the suggested workaround as mentioned in the issue, thank you!
Code is a bit hacky cause its just for testing if it works and wanted something simple, but since the features not really required theres no need yo expand on it
Codex mentioned its not really a beautiful solution and needs more work but enough to test

PR #362, opened 2026-06-08
Support generic Linux with OpenRC (Alpine, Gentoo, Artix, …)

ooonea:
## What

Adds first-class support for running cake-autorate on a **generic Linux router
with OpenRC** (Alpine, Gentoo, Artix, …), alongside the existing OpenWrt and
ASUSWRT-Merlin targets. Purely additive — the existing paths are untouched.

On OpenWrt the SQM package creates the CAKE qdiscs and cake-autorate only
*adjusts* their bandwidth. A bare Linux router has no SQM, so this adds a small
helper that brings the qdiscs up the way cake-autorate then expects.

## Changes

- **`setup.sh` / `uninstall.sh`**: detect a generic-Linux-with-OpenRC host
  (`rc-update` + `rc-service` present) → installs `/etc/init.d/cake-autorate`
  from a new OpenRC template. The GitHub-API "latest commit" lookup is made
  portable (no `jsonfilter`, which is OpenWrt-only) via `grep`/`sed`.
- **`cake-autorate.openrc.template`** (new): an `openrc-run` service — `start_pre`
  brings the qdiscs up, the launcher runs the instance(s), `stop_post` tears them
  down. It **appends** to PATH (`/usr/sbin:/sbin`) rather than overwriting, so
  OpenRC's own helpers stay reachable while `tc`/`ip`/`modprobe` resolve.
- **`sqm-setup.sh`** (new): minimal CAKE bring-up — egress CAKE on the WAN iface
  + an IFB fed by a `tc` ingress redirect, with per-host fairness
  (`nat dual-srchost` egress / `nat dual-dsthost ingress`). Interfaces + base
  rates are read from the SAME instance config (single source of truth).
- **`launcher.sh.template`**: `shopt -s nullglob` + empty-array guard, so a
  missing config set fails cleanly instead of expanding to a literal-glob path.
- **Tree is now shellcheck-clean** (`.shellcheckrc` + targeted directives); also
  fixed a `proc_pids` pinger-PID indexing bug found while linting.
- **`INSTALLATION-linux-openrc.md`** (new): install/config/run docs.

## Testing

Validated end-to-end on a real **Alpine 3.23 / OpenRC** router (Protectli N150,
Starlink WAN): the service starts at boot, the qdiscs come up, cake-autorate
adjusts the shaper under load, and `stop` tears everything down cleanly.

## Notes

`sqm-setup.sh` is only used on the generic-Linux path (OpenWrt still relies on its
SQM package). Extra CAKE keywords can be overridden per-direction via
`sqm_dl_cake_opts` / `sqm_ul_cake_opts` in the config.

### DIFF
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -21,3 +21,14 @@ enable=all
 
 # Allow shellcheck to access arbitrary files
 external-sources=true
+
+# Prefer ${var} braces (opt-in style; the codebase mixes both intentionally).
+disable=SC2250
+
+# $ / ${ } around arithmetic operands is unnecessary (cosmetic).
+disable=SC2004
+
+# Intentional array->string joins (the CPU-stat lines reuse the array name as
+# a scalar via ${arr[*]} then ${scalar// /...}); not bugs.
+disable=SC2178
+disable=SC2128
--- a/INSTALLATION-linux-openrc.md
+++ b/INSTALLATION-linux-openrc.md
@@ -1,0 +1,71 @@
+# Installation on a generic Linux router with OpenRC
+
+cake-autorate was written for OpenWrt, where the [SQM](https://openwrt.org/docs/guide-user/network/traffic-shaping/sqm)
+package creates the CAKE qdiscs and cake-autorate only *adjusts* their bandwidth.
+
+This adds first-class support for a **generic Linux router running OpenRC**
+(Alpine, Gentoo, Artix, …), which has no SQM package: a small helper,
+`sqm-setup.sh`, brings the CAKE qdiscs up, and an OpenRC service wires it all
+together. The core (`cake-autorate.sh`) is unchanged.
+
+## Requirements
+
+- `bash`, `iproute2` (`tc`), `fping`
+- kernel modules `sch_cake` and `ifb`
+- OpenRC (`rc-update`, `rc-service`)
+
+On Alpine: `apk add bash iproute2 fping`.
+
+## Install
+
+```sh
+sh -c "$(wget -qO- https://raw.githubusercontent.com/lynxthecat/cake-autorate/master/setup.sh)"
+```
+
+`setup.sh` detects OpenRC and installs into `/root/cake-autorate`, generating
+(but **not** enabling) an OpenRC service at `/etc/init.d/cake-autorate`.
+
+## Configure
+
+Edit `/root/cake-autorate/config.primary.sh`:
+
+| variable | meaning | example |
+| --- | --- | --- |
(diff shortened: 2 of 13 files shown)


lynxthecat:
@ooonea this looks well crafted but I have two concerns. 

Firstly, cake-autorate was never designed to instantiate the cake instances, rather it assumes they are set up and maintained externally. 

With this pull request we create an awkward hybrid:

- **OpenWrt (use external cake instantiation method)**
- **AsusWrt (use external cake instantiation method)**
- **OpenRC (use cake-autorate cake instantiation method)**

Might it not be better to separate out the non-OpenWrt cake instantiation method and provide it on your own GitHub repository and then we can link to that in this cake-autorate repository? So that is, users can use your mechanism to instantiate cake or other OpenWrt/AsusWrt method and then use cake-autorate for the rate changes? 

This seems more consistent to me.

And also, users might not want to be tied in to one specific way of instantiating cake. In my experience there are many different use cases for cake and I think it'd better to keep the tools: instantiation and rate changes separate for this reason. 

Personally even using OpenWrt I use my own script rather than sqm-scripts to achieve this:

https://github.com/lynxthecat/cake-qos-simple

because I wanted to support the use of DSCPs for tinning in cake, which is especially useful for lower bandwidth connections where it matters which packets are dropped first - e.g. drop bulk packets before dropping any voice packets.

Secondly, it also seems this pull request does not just add in cake instantiation capability for OpenRC (and not OpenWrt/AsusWrt), but also alters a few other things here and there, which I'd rather have seen separated out into a different pull request.

ooonea:
Thanks @lynxthecat, that makes complete sense — keeping instantiation and rate
adjustment as separate tools is the cleaner design, and I agree cake-autorate
shouldn't be the thing that brings the qdiscs up.

I've split this into two:

1. **The cake instantiation lives in its own repo now:**
   https://github.com/ooonea/cake-autorate-openrc
   It's the OpenRC/generic-Linux equivalent of what SQM does on OpenWrt (a
   `sqm-setup.sh` that creates the qdiscs cake-autorate then adjusts, plus an
   OpenRC service). cake-autorate stays completely instantiation-agnostic;
   users on a bare Linux router point it at that (or any other mechanism) and
   use cake-autorate purely for the rate changes. Happy for you to link to it
   from here if you think it's useful — and of course no obligation to.

2. **The unrelated changes are now a separate PR:** #365
   Just the shellcheck cleanup, the multi-pinger PID fix, and the launcher
   empty-config guard — nothing OpenRC.

Closing this one in favour of those. Thanks for the thoughtful review!

PR #365, opened 2026-06-20
Fix multi-pinger PID + launcher empty-config guard + shellcheck cleanup

ooonea:
## What

Two small, self-contained improvements, split out of #362 as you asked (the
parts that were unrelated to the OpenRC work).

## Changes

- **`fix`: correct pinger PID tracking in multi-pinger `ping` mode** — in the
  round-robin `ping)` method each reflector gets its own pinger indexed by
  `${pinger}`, but `proc_pids` recorded `pinger_pids[0]` for every reflector
  instead of `pinger_pids[pinger]`, so all but the first pinger were tracked
  under the wrong PID.
- **`fix`: launcher exits cleanly when no instance config is found** —
  `shopt -s nullglob` so an unmatched `config.*.sh` glob expands to nothing
  instead of a literal pattern, plus a guard that reports the missing config and
  exits non-zero rather than launching an instance with a bogus path.
- **`chore`: shellcheck-clean the tree** — quote `SC2184` unset targets, `_` for
  a throwaway `read` var, `[*]` for the intentional array→string joins on the
  CPU-stat lines, targeted `SC2155`/`SC2206` directives, and a few file-level
  disables in `.shellcheckrc`. No behaviour change.

Three atomic commits; the core loop is untouched.

### DIFF
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -21,3 +21,14 @@ enable=all
 
 # Allow shellcheck to access arbitrary files
 external-sources=true
+
+# Prefer ${var} braces (opt-in style; the codebase mixes both intentionally).
+disable=SC2250
+
+# $ / ${ } around arithmetic operands is unnecessary (cosmetic).
+disable=SC2004
+
+# Intentional array->string joins (the CPU-stat lines reuse the array name as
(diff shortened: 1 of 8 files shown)


lynxthecat:
Thanks @ooonea. I checked and all seems fine to me so I will merge this.

PR #366, opened 2026-06-21
lib.sh: fix terminate() negative-pgid handling + Fisher-Yates bias

ooonea:
Two small, independent correctness fixes in `lib.sh` (kept as two commits).

### `terminate()`: pass `--` before (negative) pids
For `pinger_method=irtt` the pinger pids are negated pgids (`set -m`), so
`kill "${pids[@]}"` sees a leading `-<pgid>` and parses it as a **signal spec**
(`invalid signal specification`), aborting the call and delivering nothing. The
graceful `TERM` on every reflector stop/replace and on cleanup therefore no-ops,
and with more than one irtt pinger the `-9` fallback aborts the same way, leaking
the `gawk`/`irtt` children across restarts. Fixed with `kill -TERM -- …` /
`kill -KILL -- …` so the negative pgids are treated as targets.

### `randomize_array()`: Sattolo's → Fisher-Yates
The function is documented as Fisher-Yates but uses `RANDOM%set` (j in
`[0, set-1]`, excluding the current index) — that is Sattolo's algorithm: only
cyclic permutations, non-uniform (an element can never stay put). `RANDOM%(set+1)`
gives the inclusive `[0, set]` an unbiased shuffle needs. Low impact (reflectors
are interchangeable and runtime swapping dilutes the startup bias), but it
contradicts the stated intent.

`shellcheck -x` and `bash -n` clean.

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -114,7 +114,11 @@ randomize_array()
 
 	for ((set=${#array[@]}-1; set>0; set--))
 	do
-		idx=$((RANDOM%set))
+		# j must be uniform in [0, set] INCLUSIVE for an unbiased Fisher-Yates.
+		# RANDOM%set gives [0, set-1], excluding the current index -> that is
+		# Sattolo's algorithm (only cyclic permutations; an element can never
+		# stay put), which biases the startup reflector shuffle.
+		idx=$((RANDOM%(set+1)))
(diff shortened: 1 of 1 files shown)


lynxthecat:
Thanks again @ooonea. All changes look good to me.

PR #367, opened 2026-06-21
cake-autorate.sh: define set_min_shaper_rates + irtt to_us/deps fixes

ooonea:
Three independent correctness fixes in `cake-autorate.sh` (separate commits; the
first is **not** irtt-specific and can be cherry-picked on its own).

### `set_min_shaper_rates` is called but undefined (any pinger method)
`f3f20a0` deleted the function and inlined it on the IDLE path, but left the
global-ping-response-timeout (stall) site calling the now-undefined name. With
`min_shaper_rates_enforcement=1`, a stall makes bash emit `command not found`,
which `intercept_stderr` turns into `kill $$` — **killing the daemon exactly when
the rate clamp was wanted**. Restored as one function used by both the stall and
IDLE sites (so the two can't drift again).

### irtt `to_us()`: bare `/s$/` shadows µs/ns
`"5µs"` and `"5ns"` also end in `s`, so the earlier `/s$/` branch swallowed them
(`mult=1e6`) and the µs/ns branches were unreachable → any sub-millisecond OWD
inflated **10⁶×**, corrupting the irtt baselines/deltas. Test bare `/s$/` last.

### irtt dependency pre-flight
The irtt path needs `gawk`, the gawk-only `time` extension and `stdbuf`, but only
`irtt` was checked. When they're missing the gawk pinger dies with stderr to
`/dev/null`, so the daemon is **not** killed — it silently produces no data and
pins CAKE at the min rate. Now names the missing dependency and exits.

`bash -n` clean; `shellcheck -x` adds no findings (−6 vs base, from de-duplicating
the IDLE inline into the restored function).

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -504,10 +504,13 @@ start_pinger()
 
 				function to_us(val, mult)
 				{
+					# Test bare /s$/ LAST: "5µs" and "5ns" also end in "s", so an
+					# earlier /s$/ branch would swallow them (mult=1000000) and the
+					# µs/ns branches would be unreachable -> sub-ms OWDs inflated 10^6x.
 					mult = (val ~ /ms$/) ? 1000    : \
-						   (val ~ /s$/)  ? 1000000 : \
 						   (val ~ /µs$/) ? 1       : \
-						   (val ~ /ns$/) ? 0.001   : -1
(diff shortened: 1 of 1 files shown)


lynxthecat:
Thanks again @ooonea. Looks good to me and I will merge.

PR #368, opened 2026-06-21
Installer robustness: setup.sh exec-after-rm + uninstall.sh CONFIG_PREFIX guard

ooonea:
Two small installer hygiene fixes (one per file, one commit each).

### `setup.sh`: don't `rm -rf "${tmp}"` right before `exec`'ing it
The non-self-replacing restart branch removed `${tmp}` and then
`exec "${tmp}/setup.sh"`, so the `exec` always failed with ENOENT and aborted the
restart, leaving a half-finished install. Dropped the premature `rm`; the
`EXIT/INT/TERM` trap still removes `${tmp}` if the `exec` fails. (Narrow reach:
only downloads predating the `SELF_REPLACING` marker take this branch.)

### `uninstall.sh`: guard `cd "${CONFIG_PREFIX}"`
Under `set -eu`, `cd "${CONFIG_PREFIX}"` aborted the whole uninstall when the
config dir didn't exist (configs already removed, a re-run after a partial
uninstall, or a Merlin/custom prefix that diverges from `SCRIPT_PREFIX`), leaving
`cake-autorate.sh`/`lib.sh`/etc. on disk. Guarded so program-file removal still
runs.

`sh -n` / `shellcheck -x` clean (the pre-existing SC2310 infos in `setup.sh` are
unrelated and untouched).

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -174,7 +174,9 @@ main() {
 				exec "${tmp}/setup.sh" "${REPOSITORY}" "${BRANCH}"
 		else
 			printf "Self-replacing not fully supported. Restarting with the new setup.sh...\n"
-			rm -rf "${tmp}"
+			# Do NOT rm -rf "${tmp}" here: exec immediately runs ${tmp}/setup.sh,
+			# so deleting it first makes the exec fail with ENOENT and aborts the
+			# restart. The EXIT/INT/TERM trap removes ${tmp} if the exec fails.
 			exec "${tmp}/setup.sh" "${REPOSITORY}" "${BRANCH}"
 		fi
 
--- a/setup.sh
+++ b/setup.sh
@@ -172,15 +172,14 @@ main() {
 			__CAKE_AUTORATE_SETUP_SH_EXEC_TMP="${tmp}" \
 			__CAKE_AUTORATE_SETUP_SH_EXEC_COMMIT="${commit}" \
 				exec "${tmp}/setup.sh" "${REPOSITORY}" "${BRANCH}"
-		else
-			printf "Self-replacing not fully supported. Restarting with the new setup.sh...\n"
-			# Do NOT rm -rf "${tmp}" here: exec immediately runs ${tmp}/setup.sh,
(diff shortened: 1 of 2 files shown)


lynxthecat:
@rany2 does this look OK to merge to you? @ooonea's other PRs seemed good to me.

rany2:
@lynxthecat It seems fine but for setup.sh I would drop self-replacing support entirely. For uninstall.sh I think it would be better to guard the entire thing with `if [ -d ${CONFIG_PREFIX} ]` same with every block that does a cd.

lynxthecat:
Thoughts @ooonea?

ooonea:
@rany2 thanks for the review — both addressed in the latest commit.

**uninstall.sh** — agreed, that's cleaner than my partial fix. Every `cd` is now guarded with `[ -d … ]` (both `CONFIG_PREFIX` and `SCRIPT_PREFIX`); if a prefix is absent there's nothing to remove there, so the block is skipped rather than aborting the rest of the uninstall under `set -eu`. I also folded the two consecutive `cd "${SCRIPT_PREFIX}"` blocks into one (the second `cd` was redundant).

**setup.sh** — I'd rather not rip out self-replacing wholesale in a robustness PR: the `…SUPPORT_SELF_REPLACING` handshake is deliberate, and removing it is a design call that's yours to make (ideally its own PR). But the *only* broken part is the `else` branch, and it's effectively dead — it fires only when the **downloaded** setup.sh lacks the marker, i.e. when installing a pre-marker (pre-2024) version. So instead of fixing the rm-then-exec there, I've **deleted the dead `else` branch**: a marker-less download now falls through and installs the downloaded files with the current setup.sh, rather than exec'ing an ancient one. The modern self-replacing path (marker present → exec) is untouched.

Happy to remove self-replacing entirely as a separate PR if you'd both prefer — just say the word.

lynxthecat:
@rany2 how about now?

lynxthecat:
Can we merge this @rany2?

rany2:
Let's not, seems too messy.

PR #369, opened 2026-06-22
Improve detection of a running cake-autorate instance

rany2:
Fixes: https://github.com/lynxthecat/cake-autorate/issues/364

PR #370, opened 2026-06-22
fix: generate_run_token uses od, not hexdump (avoid non-Essential bsdextrautils)

ooonea:
Follow-up to #369, which added `generate_run_token()`.

`generate_run_token()` (lib.sh) pipes `/dev/urandom` through **`hexdump`**:

```sh
run_token=$(head -c 32 /dev/urandom 2>/dev/null | hexdump -v -e '/1 "%02x"') || return 1
```

`hexdump` is **not** part of coreutils. On Debian it ships in the *non-Essential* `bsdextrautils` package, so on a minimal Debian box it can be absent. With `set -o pipefail` the missing binary makes the pipeline return non-zero *and* fails the `[[ ${#run_token} -eq 64 ]]` check, so `generate_run_token` returns 1. The call site is **unconditional** (every startup, any `pinger_method`) and treats that as fatal:

```
Failed to generate CAKE_AUTORATE_RUN_TOKEN. Exiting now.
```

→ the daemon never starts on a minimal Debian install.

**Fix:** use `od` (coreutils — present on Debian *and* busybox/OpenWrt):

```sh
run_token=$(head -c 32 /dev/urandom 2>/dev/null | od -An -v -tx1 | tr -d ' \\n') || return 1
```

Byte-identical output (64 lowercase hex chars) verified. Strict portability improvement, no behaviour change. `shellcheck`/`bash -n` clean.

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -129,7 +129,7 @@ generate_run_token()
 {
 	local run_token
 
-	run_token=$(head -c 32 /dev/urandom 2>/dev/null | hexdump -v -e '/1 "%02x"') || return 1
+	run_token=$(head -c 32 /dev/urandom 2>/dev/null | od -An -v -tx1 | tr -d ' \n') || return 1
 	[[ ${#run_token} -eq 64 ]] || return 1
 
 	printf '%s\n' "${run_token}"


rany2:
This would break for OpenWrt and most busybox based systems. On those systems, hexdump is available but not od. I will fix this in a different way for extremely minimal Debian systems.

ooonea:
You're right — I was solving for minimal Debian and overlooked that OpenWrt's busybox ships `hexdump` but not `od`. Thanks for catching it, and sorry for the noise.

If it's useful for the both-portable version: `cat /proc/sys/kernel/random/uuid` (twice, dashes stripped → 64 hex chars) needs neither `od` nor `hexdump` and is present on Debian and OpenWrt alike. But you said you've got a cleaner approach in mind — happy to leave it with you.

PR #374, opened 2026-06-22
harden: restrict the runtime dir to mode 0700

ooonea:
`mkdir -p "${run_path}"` creates the root-owned runtime directory with the process umask (typically `022` → world-readable). `${run_path}` holds `proc_pids`/`run_token` and the generated `log_file_export`/`log_file_reset` helper scripts, and the code already flags the directory as sensitive (the `mkdir -p ${run_path}`/`rm -r ${run_path}` warning banner). Least-privilege for a root-owned runtime dir: create it, then `chmod 0700`.

`chmod` after `mkdir` rather than `mkdir -m 0700` to avoid `shellcheck` SC2174 (`-m` with `-p` only applies to the deepest dir) and to guarantee the mode on the final directory.

No behaviour change for the daemon (runs as root). `bash -n` / `shellcheck` clean.

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -1013,12 +1013,10 @@ then
 	else
 		log_msg "DEBUG" "${run_path} already exists but no conflicting instance is running. Removing and recreating."
 		rm -r "${run_path}"
-		mkdir -p "${run_path}"
-		chmod 0700 "${run_path}"
+		( umask 077 && mkdir -p "${run_path}" )
 	fi
 else
-	mkdir -p "${run_path}"
-	chmod 0700 "${run_path}"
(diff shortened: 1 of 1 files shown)


rany2:
I think umask would be nicer than this.

ooonea:
Good call — switched to `( umask 077 && mkdir -p "${run_path}" )`. Cleaner than the trailing chmod (no SC2174, no extra syscall), and it also covers any intermediate directory `mkdir -p` creates, not just the deepest one.

lynxthecat:
Hey @rany2 if this is OK with you now then please merge. 

And thanks again @ooonea.

PR #376, opened 2026-06-22
fix: clear the log-export result marker before signalling (race)

ooonea:
The generated `log_file_export` helper (`generate_log_file_scripts`) does:

```sh
if kill -USR1 "${maintain_log_file_pid}"   # signal first
then …
fi
rm -f "${run_path}/last_log_file_export"   # then remove the previous marker
while [[ ! -f "${run_path}/last_log_file_export" ]]   # then wait for a fresh one
do sleep 1; … timeout … done
```

The daemon's `export_log_file` writes `last_log_file_export` on USR1. If it runs **between** the `kill` and the `rm`, the helper deletes the just-written marker, never sees a fresh one, and exits 1 with a spurious "Timeout reached" even though the export completed on disk.

**Fix:** clear the marker *before* signalling. One-line reorder. (`log_file_reset` has no marker wait, so it is unaffected.)

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -260,6 +260,12 @@ generate_log_file_scripts()
 
 	timeout_s=\${1:-20}
 
+	# Clear the previous result marker BEFORE signalling. The daemon's
+	# export_log_file writes this marker; removing it *after* the signal can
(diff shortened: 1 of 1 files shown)


lynxthecat:
Yes, nice catch.

PR #383, opened 2026-06-24
fix: keep a monitor-only direction (adjust_*_shaper_rate=0) pinned at base (#377)

### LINKED ISSUE #377: adjust_*_shaper_rate=0 with min<max: internal shaper rate drifts off the real fixed rate, corrupting load_percent
Filing as an issue rather than a PR because the fix involves a small semantics decision I'd rather confirm with you first.

### Problem
With `adjust_dl_shaper_rate=0` (compute but don't tc the download rate), the internal `shaper_rate_kbps[dl]` is still driven by the full rate-update algorithm:

- it's initialised to `base_dl_shaper_rate_kbps` (`cake-autorate.sh:1232`), not the real out-of-band fixed rate;
- the rate-update case (`:1655-1733`) updates `shaper_rate_kbps[dir]` for **both** directions unconditionally; `adjust_shaper_rate[dir]` gates only the `tc` call in `set_shaper_rate` (`:743-748`);
- it's then clamped to `[min,max]` (`:1735-1739`).

So when `min_dl < max_dl`, the internal DL rate floats within `[min,max]` **decoupled from the real (un-applied) fixed rate**, and `load_percent[dl]=100*achieved/shaper_rate_kbps[dl]` (`:1643-1644`) is computed against that phantom rate. Because the EWMA-delta gate is `(( load_percent[dl] < high_load_thr && load_percent[ul] < high_load_thr ))` in every pinger method (`:1457` and siblings), a wrong `load_percent[dl]` **freezes the UL EWMA update**, degrading UL reflector selection — i.e. a phantom load on the *unmanaged* direction corrupts the *managed* one.

### When it bites
A user doing a static download cap the natural way — `base_dl=max_dl=cap`, `adjust_dl=0`, leaving `min_dl` at the default — hits it: on any detected DL 'bufferbloat' the internal rate decays toward `min_dl` while the real rate stays at `cap`. It's **avoided** by setting `min_dl=base_dl=max_dl` (the clamp then pins the internal rate to the real one), which is why a tight static cap doesn't show it.

### Fix options (your call)
1. When `adjust_shaper_rate[dir]==0`, pin `shaper_rate_kbps[dir]` to `base_shaper_rate_kbps[dir]` (skip driving it) and document that `base_*` MUST equal the real fixed rate when the matching `adjust_*=0`.
2. Compute `load_percent[dir]` against `base_shaper_rate_kbps[dir]` when `adjust=0`.

Both change the documented meaning of `base_*` under `adjust=0`, so I wanted to ask which framing you prefer before sending a patch. (Context: static-cap-on-Starlink is exactly this configuration.)

ooonea:
A direction with `adjust_${direction}_shaper_rate=0` is monitor-only: `set_shaper_rate()` never issues a `tc` change for it, but its internal `shaper_rate_kbps` still walks the `[min, max]` range through the rate-update arms, the min/max clamp and `set_min_shaper_rates()`. Since `load_percent` is computed against `shaper_rate_kbps`, the reported load for the unmanaged direction is referenced to a phantom rate instead of the real, fixed CAKE bandwidth — with a fixed 200 Mbit/s download shaper and the internal rate drifted to the shipped default maximum, an empirically measured true 50% load was reported as 125%.

**v2, per review feedback: fix it in data, not control flow.** At startup, a monitor-only direction's *runtime* bounds collapse onto base (`min_shaper_rate_kbps[dir] = max_shaper_rate_kbps[dir] = base`). Every existing `shaper_rate_kbps` mutation site then lands the rate back on base:

- the min/max clamp is **untouched** — still one arithmetic block, byte-identical to master (this addresses the review comment favouring merged arithmetic blocks on the frequently-run path);
- the rate-update `for direction in dl ul` loop and its case arms are untouched;
- `set_min_shaper_rates()` now reads the runtime bounds arrays instead of the config scalars (cold path, idle/stall only; behaviour-identical for driven directions), so it too pins a monitor-only direction to base.

The config scalars are not modified: the startup config dump still shows what the user wrote, and the `min <= base <= max` startup validation (#387) runs on the scalars before this point and is unaffected.

A monitor-only direction's `base_${direction}_shaper_rate_kbps` must match the CAKE bandwidth actually fixed on the interface for `load_percent` to be meaningful; a one-shot DEBUG note records this at startup for log forensics.

Side effect worth noting: with the internal rate pinned, `set_shaper_rate()`'s early return silences the misleading SHAPER log lines master currently emits for a direction it never actually `tc`-changes.

`bash -n` clean; `shellcheck` identical to master (the 2 pre-existing SC2310 info notes, nothing new).

Fixes #377.

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -792,7 +792,7 @@ set_min_shaper_rates()
 	# calling the now-undefined name -> command-not-found -> intercept_stderr
 	# kills the daemon exactly when the clamp was wanted. Restore it as one
 	# function used by both sites.
-	shaper_rate_kbps[dl]=${min_dl_shaper_rate_kbps} shaper_rate_kbps[ul]=${min_ul_shaper_rate_kbps}
+	shaper_rate_kbps[dl]=${min_shaper_rate_kbps[dl]} shaper_rate_kbps[ul]=${min_shaper_rate_kbps[ul]}
 	set_shaper_rate "dl"
 	set_shaper_rate "ul"
 }
@@ -1294,6 +1294,25 @@ last_shaper_rate_kbps[dl]=0 last_shaper_rate_kbps[ul]=0 \
 interface[dl]=${dl_if} interface[ul]=${ul_if} \
 adjust_shaper_rate[dl]=${adjust_dl_shaper_rate} adjust_shaper_rate[ul]=${adjust_ul_shaper_rate} \
 dl_max_wire_packet_size_bits=0 ul_max_wire_packet_size_bits=0
+
+# A monitor-only direction (adjust_${direction}_shaper_rate=0) never has its
+# CAKE shaper changed, but its internal shaper_rate_kbps still divides into
+# load_percent. Collapse its runtime bounds onto base so that every existing
+# shaper_rate_kbps mutation site (the rate-update arms, the min/max clamp and
+# set_min_shaper_rates) lands the rate back on base, keeping load_percent
+# referenced to the real fixed rate without touching the rate-update hot path
+# (issue #377). base must therefore match the CAKE bandwidth actually fixed on
+# the interface -- noted at DEBUG for log forensics.
+if ((adjust_dl_shaper_rate == 0))
+then
+	min_shaper_rate_kbps[dl]=${base_dl_shaper_rate_kbps} max_shaper_rate_kbps[dl]=${base_dl_shaper_rate_kbps}
+	log_msg "DEBUG" "adjust_dl_shaper_rate=0 -- dl is monitor-only; load_percent[dl] is referenced to base_dl_shaper_rate_kbps (${base_dl_shaper_rate_kbps}), which should match the fixed download CAKE bandwidth on ${dl_if}."
+fi
+if ((adjust_ul_shaper_rate == 0))
+then
+	min_shaper_rate_kbps[ul]=${base_ul_shaper_rate_kbps} max_shaper_rate_kbps[ul]=${base_ul_shaper_rate_kbps}
+	log_msg "DEBUG" "adjust_ul_shaper_rate=0 -- ul is monitor-only; load_percent[ul] is referenced to base_ul_shaper_rate_kbps (${base_ul_shaper_rate_kbps}), which should match the fixed upload CAKE bandwidth on ${ul_if}."
+fi
 
 get_max_wire_packet_size_bits "${dl_if}" dl_max_wire_packet_size_bits
 get_max_wire_packet_size_bits "${ul_if}" ul_max_wire_packet_size_bits


lynxthecat:
@ooonea do you see what I mean with my question above?

rany2:
@lynxthecat this seems to make sense, why not merge it?

ooonea:
@lynxthecat sorry for the slow reply. I want to make sure I'm answering the right thing, because the only open question of yours I can find is the one from #377 — avoiding any extra computation per rate-update iteration for the common both-directions case. If that's the one, here's how the PR handles it:

- The driven-directions decision is made **once, at startup**: `update_directions` is built from `adjust_dl/ul_shaper_rate` before the main loop starts. Nothing is re-evaluated at runtime.
- With both directions driven (the typical case) the list is exactly `dl ul`, so the rate-update loop iterates the same two directions as today — no added per-iteration conditional, the only difference is expanding a two-element array instead of two literals.
- The min/max clamp keeps the same comparisons, just scoped to the driven list; `set_min_shaper_rates` only runs on the idle/stall paths, which are not hot.

If you meant a different question: I don't see any inline or review comment from you on this PR (the API shows none), so if you left one it may still be sitting in an unsubmitted "Pending" review — GitHub only shows those to their author. Happy to address it as soon as it's visible.

lynxthecat:
I meant this comment here: 

https://github.com/lynxthecat/cake-autorate/pull/383/changes#r3477127697

Namely I favour one arithmetic block rather than two because merging arithmetic blocks wherever possible gives a (tiny) reduction in computation, and the arithmetic block here is run very frequently.

ooonea:
Thanks — now I see what you meant. Amended accordingly: the clamp (and the whole rate-update path) is now **completely untouched** — still the single merged arithmetic block, byte-identical to master.

v2 fixes it in data instead of control flow: at startup, a monitor-only direction's runtime bounds collapse onto base (`min_shaper_rate_kbps[dir] = max_shaper_rate_kbps[dir] = base`), so the existing clamp itself re-pins the internal rate to base right after the rate-update arms. `set_min_shaper_rates()` now reads the runtime bounds arrays instead of the config scalars (cold path, idle/stall only; behaviour-identical for driven directions), so the idle/stall pin lands on base too.

Net result: zero changes on the frequently-run path for everyone, and the diff shrank from ~50 lines to ~20. PR body updated to describe the new shape.

lynxthecat:
Yes we're there pretty much with this PR. I'm just not sure about the initial three lines of comments because they seem misleading. To my mind we swapped out to array from scalars and so this is just a change needed to ensure we remain consistent and use the array. I'm not sure we actually need the comment at all here. But if we use one I think it could be clearer.

ooonea:
Dropped, and I agree with your reading: the comment made it sound as if the arrays were being reached for *here* because of the monitor-only case, when this line is just the swap from the config scalars to the runtime bounds arrays being carried through consistently — behaviour-identical for any direction that is actually driven. The reason the bounds can differ from the scalars in the first place is already stated where they are collapsed, which is the place that needs it.

Force-pushed as b3b72af: the diff is now +20/-1, the one-line swap in `set_min_shaper_rates()` plus the startup block. `bash -n` clean and `shellcheck -x` matches master's baseline exactly (3x SC2310, none new).

lynxthecat:
Thanks @ooonea.

PR #384, opened 2026-06-29
Add support for Debian/Ubuntu

sbivol:
This adds support for running cake-autorate on Debian and related distributions.

I have been running cake-autorate on Ubuntu Server for a year and am willing to maintain support for this OS family going forward.

### DIFF
--- a/cake-autorate@.service
+++ b/cake-autorate@.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=cake-autorate for %i
+BindsTo=sys-subsystem-net-devices-%i.device
+After=sys-subsystem-net-devices-%i.device
+
+[Service]
+Environment="CAKE_AUTORATE_SCRIPT_PREFIX=/opt/cake-autorate"
+ExecStartPre=/opt/cake-autorate/ifb-setup.sh %i
+ExecStart=/opt/cake-autorate/launcher.sh
+Restart=on-failure
+RestartMode=direct
+RestartSec=5
+TimeoutStopSec=10
+StandardOutput=journal
+
+[Install]
+WantedBy=multi-user.target
--- a/cake-autorate@.service
+++ b/cake-autorate@.service
@@ -1,0 +1,28 @@
+[Unit]
+Description=cake-autorate for %I
+BindsTo=sys-subsystem-net-devices-%i.device
+After=sys-subsystem-net-devices-%i.device
+
+[Service]
+AmbientCapabilities=CAP_NET_ADMIN
+AmbientCapabilities=CAP_NET_RAW
+CapabilityBoundingSet=CAP_NET_ADMIN
+CapabilityBoundingSet=CAP_NET_RAW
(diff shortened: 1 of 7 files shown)


lynxthecat:
Nice to add support for more use cases, but cake-autorate itself should not instantiate cake instances. That should be taken care of independently of cake-autorate. cake-autorate should just change the bandwidth of existing cake instances.

sbivol:
The challenge is in there not being an easy way to get _cake_ set up outside router-oriented distributions.
I can move `ifb-setup.sh` to a separate repo and reference it in the documentation.
What are your thoughts?

lynxthecat:
Sounds good. I think that was also just recently done for another such PR.

lynxthecat:
@sbivol shall I keep this PR open?

cake-autorate@.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=cake-autorate for %i
```
rany2: It should be %I not %i in case an interface contains hyphens. systemd requires that any service with hyphen be escaped by `\x2d`. %I unescapes `\x2d` so that we get the proper interface name

rany2 (changes requested):

sbivol:
Please keep this PR open, I have split `ifb-setup.sh` into its own repo at https://github.com/sbivol/cake-on-wan and am still testing it.
The next steps are to add documentation to that repo and update this PR accordingly.

sbivol:
The changes are now in place, please review again.

rany2:
This is what I am using:

```ini
[Unit]
Description=cake-autorate for %I
BindsTo=sys-subsystem-net-devices-%i.device
After=sys-subsystem-net-devices-%i.device

[Service]
AmbientCapabilities=CAP_NET_ADMIN
AmbientCapabilities=CAP_NET_RAW
CapabilityBoundingSet=CAP_NET_ADMIN
CapabilityBoundingSet=CAP_NET_RAW
ExecStart=%%PREFIX/cake-autorate %%CONFIG/config.%I.sh
NoNewPrivileges=true
PrivateTmp=true
ProtectHome=true
ProtectSystem=strict
Restart=on-failure
RestartSec=5
RestrictAddressFamilies=AF_INET
RestrictAddressFamilies=AF_INET6
RestrictAddressFamilies=AF_NETLINK
RestrictAddressFamilies=AF_UNIX
RuntimeDirectory=cake-autorate

[Install]
WantedBy=multi-user.target
```

I would avoid using the launcher.

rany2:
I also added https://github.com/lynxthecat/cake-autorate/commit/d2ff9aa4c090784917ec165c61f3f5c2e4841c66 which should better accommodate systemd.

sbivol:
Updated the branch with expanded documentation, executing `cake-autorate.sh` instead of the launcher, and additional runtime restrictions from https://github.com/lynxthecat/cake-autorate/pull/384#issuecomment-5318159773.

rany2:
Thank you!

PR #386, opened 2026-07-19
startup: check for a conflicting instance before rotating the log

ooonea:
Split out from the closed #373, which conflated two unrelated fixes. This is the log-rotation half; the config-bounds half is a separate PR.

### The bug

The running-instance check (present before #369, which strengthened it with the run-token validation) runs *after* `rotate_log_file`. So starting a redundant second instance of an already-running config rotates/truncates the shared log file first, and only then detects the conflict and exits — corrupting the log of the instance that is legitimately running.

### The fix

Move the `run_path` conflicting-instance check ahead of `rotate_log_file` (and ahead of the stderr-intercept setup and the "Starting" SYSLOG line). A redundant start now exits without touching the running instance's log or spawning `intercept_stderr`; the log is only rotated once we know we are actually going to start.

Pure reordering — the check, the rotate and the redirect blocks are unchanged, only their order is. `bash -n` clean; `shellcheck -x`: same findings as `master` (only their line numbers shift with the move).

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -996,18 +996,11 @@ else
 	log_file_path="/var/log/cake-autorate${instance_id:+.${instance_id}}.log"
 fi
 
-rotate_log_file
-
-# save stderr fd, redirect stderr to intercept_stderr
-# intercept_stderr sends stderr to log_msg and exits cake-autorate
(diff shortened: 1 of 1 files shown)


lynxthecat:
Nice improvement. Thanks for this @ooonea.

Issue #379, opened 2026-06-22
mqtt-publisher: several correctness/availability issues + credential on argv


ooonea:
A review pass over `mqtt-publisher.sh` (and its uninstall path) surfaced the items below. Grouping them in one issue since they're one subsystem — happy to send PRs for whichever you/@bernhardberger want fixed (or you may prefer to take them directly). Line refs against master `27356f4`.

**SECURITY — broker password on the command line** (`:30,50,118,203`)
The password is passed as `-P "$MQTT_PASS"` at every `mosquitto_pub` call, so it lands in world-readable `/proc/<pid>/cmdline` (`ps -ef`); the long-lived `-l` pipeline keeps it continuously exposed. The mosquitto man page documents `-P` as insecure. Suggest `mosquitto_pub --pw-file <0600 file>` (mosquitto ≥2) or stdin.

**Publisher is mute with the shipped config** (`mqtt-publisher.sh:143,158` + `defaults.sh:22,24`)
The data path keys off `SUMMARY`/`CPU` log records, both off by default (`output_summary_stats=0`, `output_cpu_stats=0`); README even advises disabling `output_summary_stats`. With the shipped config the publisher emits discovery + retained `online` but the sensors stay empty forever, undocumented. Suggest documenting the `output_summary_stats=1` prerequisite and warning loudly when no enabled config sets it.

**CPU core count source mismatch** (`:4,161`)
`CPU_CORES=$(grep -c '^processor' /proc/cpuinfo)` while cake-autorate emits `cpu_cores` from `cpu[0-9]` lines in `/proc/stat`; the in-awk `if (num_cores==0) num_cores=NF-5` fallback is dead (cpuinfo always ≥1). If the two diverge (offline/unplugged core) the awk reads past `NF` → phantom 0.0% cores or dropped cores.

**Stale categoricals on a CPU-only emit window** (`:151-154,168,191-197`)
`event_epoch=max(summary_epoch,cpu_epoch)` can fire on a CPU-only window (`cn>0,sn==0`); the reset zeroes numeric accumulators but not the latest-wins `dl_load`/`ul_load`/`cake_dl`/`cake_ul`, so the payload mixes `*_rate_kbps:0.0`+`samples:0` with a stale `dl_load_condition`/`cake_dl_rate_kbps`. Suggest gating the SUMMARY-derived emit on `sn>0`.

**LWT not published on a clean pipeline exit** (`:116-121,201-208`)
[...]

PR #390, opened 2026-07-26
mqtt-publisher: keep the broker password off the command line (#379)

ooonea:
Addresses the security item from #379: the broker password was passed as `-P "$MQTT_PASS"` at all four `mosquitto_pub` call sites, so it sat in `/proc/<pid>/cmdline` for any process on the box to read. The streaming publisher uses `-l` and is long-lived, so the exposure was continuous rather than momentary.

### Mechanism

Neither of the two options named in #379 works here, as discussed in [this comment](https://github.com/lynxthecat/cake-autorate/issues/379#issuecomment-5083386625): `--pw-file` has never existed in any mosquitto client, and `-o` needs 2.1.0 while OpenWrt ships 2.0.22. What 2.0.22 does have is the client's own default config file — `client_config_load()` reads `$XDG_CONFIG_HOME/mosquitto_pub` and parses its lines with the same option parser as argv (`client/client_shared.c:291-321`, with `-P`/`--pw` at :862 and `-u`/`--username` at :1097).

So: `mktemp -d` (0700), the two options written under `umask 077`, and the four call sites prefixed with `XDG_CONFIG_HOME=` instead of carrying the credentials. `cleanup()` removes the directory *after* the shutdown `offline` publishes, which still need to authenticate. No new dependency, no version floor above what OpenWrt already ships.

### Behaviour change worth naming

With `MQTT_USER` and `MQTT_PASS` both empty, no credential file is written and neither flag is passed, where previously two empty strings were. Anonymous brokers should be unaffected in practice, but it is a real difference and I would rather flag it than have it found later.

### Verification

Against a local broker with `allow_anonymous false` and a password file, running this actual script against a stub deployment (config with `log_file_path_override`, one `cake-autorate.*.log`):

- the publisher authenticates and the retained `cake-autorate/test/availability online` arrives — so the credentials are being read from the file
- `/proc/<pid>/cmdline` of the live `-l` publisher: `mosquitto_pub -h 127.0.0.1 -p 18831 -t cake-autorate/test -l -q 1 --will-topic … --will-retain` — no password
- the credential file lands `0600`
- after shutdown the directory is gone

`bash -n` clean; `shellcheck -S style` reports only the two pre-existing `SC1091` source-not-followed infos.

### Scope

This is only the argv half of group 1 from #379. The clean-exit `offline` publish and group 3 remain with @bernhardberger, per the split agreed in that thread.

One thing this does *not* fix, noted in #379: `MQTT_PASS` is still a literal in `mqtt-publisher.sh`, which installs `0755`. Getting it off argv removes the continuous `ps` exposure, but the secret living in a world-readable script is a separate question for whoever owns the install path.

### DIFF
--- a/mqtt-publisher.sh
+++ b/mqtt-publisher.sh
@@ -26,9 +26,8 @@ cleanup()
     for log_dir in "${log_dirs[@]}"; do
         for log_file_path in "${log_dir}"/cake-autorate.*.log; do
             local instance="$(basename "${log_file_path}" | sed -E 's/^cake-autorate\.([^.]+)\.log$/\1/')"
-            mosquitto_pub \
+            XDG_CONFIG_HOME="${mqtt_config_dir}" mosquitto_pub \
                 -h "$MQTT_HOST" -p "$MQTT_PORT" \
-                -u "$MQTT_USER" -P "$MQTT_PASS" \
                 -r -q 1 \
                 -t "${BASE_MQTT_TOPIC}/${instance}/availability" \
                 -m "offline" 2>/dev/null || true
@@ -38,6 +37,8 @@ cleanup()
     for pid in "${publish_stats_pids[@]}"; do
         kill -- -"$pid" 2>/dev/null || true
     done
+    # after the shutdown publishes above, which still need to authenticate
+    rm -rf "${mqtt_config_dir}"
     exit 0
 }
 
@@ -47,9 +48,8 @@ publish_config()
     local object_id="$2"
     local payload="$3"
 
-    mosquitto_pub \
+    XDG_CONFIG_HOME="${mqtt_config_dir}" mosquitto_pub \
         -h "$MQTT_HOST" -p "$MQTT_PORT" \
-        -u "$MQTT_USER" -P "$MQTT_PASS" \
         -r -q 1 \
         -t "$DISC_PREFIX/sensor/$device_id/$object_id/config" \
         -m "$payload"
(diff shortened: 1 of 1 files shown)


lynxthecat:
Does this use `/var/run/cake-autorate/test/`? Why not just `/var/run/cake-autorate/`? How about shortening the file to `active`?

ooonea:
That one is an MQTT topic rather than a path — sorry, the verification section reads like a directory. `cake-autorate/test/availability` is where the retained availability message is published, and `test` is just the instance name of the stub config I ran against.

Nothing in this PR writes under `/var/run`. The credential file goes into a `mktemp -d` directory that `cleanup()` removes, and it has to be named `mosquitto_pub`, because that is the filename the client's own `client_config_load()` looks for under `$XDG_CONFIG_HOME`. The publisher also isn't tied to a single instance — it discovers them from the log files and fans out — so it has no one run directory to put this in, whereas a private directory it owns for its own lifetime needs no coordination with the daemons.

The topic layout itself is pre-existing and untouched here: `BASE_MQTT_TOPIC="cake-autorate"`, with `${BASE_MQTT_TOPIC}/${instance}` for the stats and `${BASE_MQTT_TOPIC}/${instance}/availability` for the birth/LWT message. The `${instance}` level is what keeps two cake-autorate instances on one box from colliding, and the Home Assistant discovery payloads point at it as `availability_topic`.

Shortening `availability` to `active` is reasonable, but I would do it in its own PR rather than fold it into this one: it is a retained topic, so existing installs would be left with a stale retained `online` on the old name. Happy to pick it up separately if you want it — I kept it out of here only because it changes the Home Assistant-facing topic layout rather than the credential handling.

lynxthecat:
Ah, got it. Thanks.

Issue #353, opened 2026-02-13
FR: implement support for cake_mq


Snuupy:
https://www.reddit.com/r/openwrt/comments/1r1muvf/cake_mq_patches_landed_in_the_openwrt2512_branch/

it's not in a snapshot available to me yet but it would enable >1gbps routing for many devices including my multi-core routers, especially on variable connections

CHC383:
I tried the `tc qdisc change` command and it does require `cake`/`cake_mq`. Not sure if there is better way to detect whether cake_mq is enabled, but one way is check the return value of `tc qdisc show | grep -q cake_mq` and choose `cake`/`cake_mq` accordingly.

lynxthecat:
@ooonea do you see a good way to address this one that's been outstanding for a while?

ooonea:
@lynxthecat your `tc` test settles it: `tc qdisc change ... bandwidth` won't parse without naming the qdisc, so we can't drop `cake` — and cake_mq needs the keyword `cake_mq` (underscore, as it appears in `tc qdisc show`: `qdisc cake_mq 801c: root ...`). @CHC383's `grep` instinct is the right one; I'd just refine it into a one-time detection at startup so there's zero per-iteration cost:

- right after we already run `tc qdisc show dev <iface>` for the wire-packet-size, detect the **root** qdisc kind per direction and store it (`cake` vs `cake_mq`), defaulting to `cake` if the output is empty/ambiguous;
- then use that stored keyword in the single `tc qdisc change` call.

Anchoring the match to the `root` line matters: cake_mq is the mq-root wrapper with plain `cake` leaves hanging off it, so a bare `grep cake` would false-match a leaf. It's ~10 lines, no config-schema change, and no behaviour change for single-core / plain-`cake` users (they keep getting `cake`). Happy to send it if you like the shape.

Aquathing:
@lynxthecat any developments on this? I can help too with testing (I'll have a Flint 2 in few days and want to install openwrt 25)

Aquathing:
Actually it looks like the latest release kinda broke cake_mq

    SQM CAKE MQ (cake_mq): throughput may be unexpectedly low on some configurations after the scheduler fixes in this release. https://github.com/openwrt/openwrt/issues/22344

lynxthecat:
Hmm yes this needs resolving. I’ll try to work on this soon.

Aquathing:
[...]

PR #391, opened 2026-07-26
shaper: detect a cake_mq root qdisc instead of assuming cake (#353)

ooonea:
Closes the gap reported in #353. @lynxthecat asked me in that thread whether I saw a good way to address it; this is the shape I proposed there, now written out.

### The failure mode

`set_shaper_rate()` hardcodes `cake` in its `tc qdisc change root` call. If the root qdisc is `cake_mq`, the kernel rejects the change with `Invalid qdisc name` — but the call sends stderr to `/dev/null` and its exit status is never checked, so cake-autorate keeps emitting `SHAPER` log lines while shaping nothing. It fails silently, which is worse than failing loudly: from the logs everything looks fine.

This is reachable today, not hypothetical: sqm-scripts picks `cake_mq` whenever `USE_MQ=1` and the device has more than one queue (`functions.sh:966-968`), and OpenWrt carries the kernel side in `backport-6.12` (`700-02`..`700-07`) plus the iproute2 side in `001-1-tc-cake-add-cake_mq-support`. Configuring the `cake_mq` instance directly and letting the config reach the sub-qdiscs is the designed behaviour — `700-04` is literally "Share config across cake_mq sub-qdiscs".

### The change

~15 lines. `get_root_qdisc_kind()` reads `tc qdisc show root dev <if>` once per direction at startup and stores the kind in a new `qdisc_kind[dl|ul]`; `set_shaper_rate()` uses that instead of the literal. Anything unrecognised — `htb`, `noqueue`, empty output, a leaf line — falls back to `cake`, so a conventional setup is byte-identical to today. No per-iteration cost, no new config knob.

### What I tested, and what I could not

Tested:

- 6/6 detection cases against real `tc` output strings: `cake_mq` root → `cake_mq`; `cake` root, `htb` root, `noqueue`, a `parent 1:1` leaf line, and empty output → `cake`
- against a **real** `tc qdisc show root`, in a network namespace with a dummy interface: a genuine `cake` root is detected as `cake`; `fq_codel` and `htb` roots fall back to `cake`
- `bash -n` clean, `shellcheck -S warning` reports nothing on the new lines

Not tested, and this is the part that needs someone with the hardware: **I have no cake_mq-capable box**, so I have not observed an actual `tc qdisc change root dev <if> cake_mq bandwidth <X>` taking effect end to end. The detection is verified; that the subsequent rate change propagates to the sub-qdiscs rests on reading `700-04`/`701-02` rather than on my having seen it work.

Marking this a draft for exactly that reason. @Snuupy, @CHC383 — you raised #353 and suggested the `grep -q cake_mq` approach; if either of you can run this on a multi-queue setup with `USE_MQ=1` and confirm the shaper actually tracks, I will flip it out of draft. If it turns out `change` on a `cake_mq` root needs different handling, better to find that now than after merge.

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -754,11 +754,11 @@ set_shaper_rate()
 
 	(( shaper_rate_kbps[${direction}] != last_shaper_rate_kbps[${direction}] )) || return
 
-	((output_cake_changes)) && log_msg "SHAPER" "tc qdisc change root dev ${interface[${direction}]} cake bandwidth ${shaper_rate_kbps[${direction}]}Kbit"
+	((output_cake_changes)) && log_msg "SHAPER" "tc qdisc change root dev ${interface[${direction}]} ${qdisc_kind[${direction}]} bandwidth ${shaper_rate_kbps[${direction}]}Kbit"
 
 	if ((adjust_shaper_rate[${direction}]))
 	then
-		tc qdisc change root dev "${interface[${direction}]}" cake bandwidth "${shaper_rate_kbps[${direction}]}Kbit" 2> /dev/null
+		tc qdisc change root dev "${interface[${direction}]}" "${qdisc_kind[${direction}]}" bandwidth "${shaper_rate_kbps[${direction}]}Kbit" 2> /dev/null
 	else
 		((output_cake_changes)) && log_msg "DEBUG" "adjust_${direction}_shaper_rate set to 0 in config, so skipping the corresponding tc qdisc change call."
 	fi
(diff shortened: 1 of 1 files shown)


lynxthecat:
This looks pretty good to me. Pedantic me wondered about `qdisc_type` rather than `qdisc_kind`. Any thoughts on which is a better fit?

ooonea:
I went with `kind` because it is iproute2's own word for this rather than a preference of mine. `tc qdisc help` ends its usage with `[ [ QDISC_KIND ] [ help | OPTIONS ] ]` and spells out `QDISC_KIND := { [p|b]fifo | tbf | prio | red | etc. }`; the parser function behind it is `get_qdisc_kind()`, and on the netlink side the attribute carrying the string is `TCA_KIND`. `tc qdisc show` then prints it immediately after the word `qdisc`, which is exactly what the regex here anchors on.

There is a local reason too: `type` is already spoken for twice in `cake-autorate.sh` — the message class in `log_msg` (`local type=${1}`) and the shell variable type in the config validation (`typeof`, `user_type`, `valid_type`). `qdisc_kind` keeps the three apart at a glance. That said it is your codebase, so if you prefer `qdisc_type` say the word and I will rename it.

Still keeping this in draft until someone with a multi-queue box can confirm the `tc qdisc change` actually lands on a `cake_mq` root — the detection is covered by fixtures and a real `tc qdisc show root` in a netns here, but I have no hardware to prove the change propagates.

lynxthecat:
Perhaps I can test on my RT3200 actually?

ooonea:
That would be great! The detection half is already covered (fixtures plus a live `tc qdisc show` parse), but whether the `change` calls behave on a real cake_mq root is exactly the part I could not verify without multi-queue hardware. If it works on your RT3200, I'll mark this ready for review.

PR #392, opened 2026-07-29
fix: reject non-numeric RTT samples (a clock step mid-ping kills the daemon)

ooonea:
On the first boot after an OS upgrade, cake-autorate died 35 seconds in and my WAN ran unshaped until I restarted it by hand. What made it worse is that systemd saw a clean exit — "Deactivated successfully" — so no `Restart=` policy would have caught it. The shaper was just quietly gone.

The daemon log and chronyd, in the same second:

```
ERROR; 2026-07-29-17:03:58; .../cake-autorate.sh: ((:
ERROR; 2026-07-29-17:03:58; dl_owd_us=10#: invalid integer constant (error token is "10#")

Jul 29 17:03:58 chronyd[1273]: System clock was stepped by -1.759986 seconds
```

The clock stepped backwards while a ping was in flight, so fping reported a negative RTT. It prints such a value rather than dropping it — `sprint_tm()` has a branch labelled `/* negative (unexpected) */` that formats it with `%.2g`. The line still has 12 fields, so it passed the response gate, and then the parse hit this:

```console
$ rtt_ms="-1.8e+03"; printf -v rtt_us %.3f "${rtt_ms}"; (( dl_owd_us=10#${rtt_us//.}/2 ))
bash: ((: dl_owd_us=10#: invalid integer constant (error token is "10#")
```

Base notation can't carry a sign, bash aborts the whole arithmetic compound, and the cleanup trap takes the daemon down with exit 0.

The fix rejects an RTT token containing anything outside `[0-9.]`, at the response gate. It's a glob class check rather than a regex on purpose: `[[ =~ ]]` recompiles the ERE on every sample, which measured ~32 µs against ~0.5 µs for the pattern match. A line that doesn't match simply isn't treated as a reflector response, so there's no new control flow, and a nonsense sample no longer reaches the OWD baselines.

Only the fping arm needs this. As @rany2 pointed out, iputils clamps a backward step to zero instead of reporting it, so `ping` can't produce a value that breaks the arithmetic. irtt, tsping and fping-ts are unaffected too — their parsers subtract integer timestamps, where a negative intermediate value is legal.

Checked with `bash -n`, and `shellcheck -x` reports the same two pre-existing findings as master and nothing new. I fed the gate normal fping lines, negative-RTT lines, `ICMP Host Unreachable` and `timed out` lines, and it accepts and rejects the right ones. It's been running on my own router since the incident.

For the record: the router is NixOS, not Debian as this description originally said — it was Debian until mid-July, and I carried the old wording over by mistake.

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -1464,7 +1464,15 @@ do
 					fi
 					;;
 				fping)
-					if ((${#command[@]} == 12))
+					# fping reports a negative RTT when the clock steps backwards
+					# mid-ping: it prints such a value rather than clamping it, and
+					# 10#${rtt_us//.} cannot represent a negative, so the arithmetic
+					# error kills the main loop. Reject any RTT token carrying a
+					# character outside [0-9.] — a glob class check, not a regex,
+					# since [[ =~ ]] recompiles the ERE on every sample.
+					# The ping arm needs no equivalent: iputils clamps a backward
+					# step to zero instead of reporting it.
+					if ((${#command[@]} == 12)) && [[ ${command[6]} != *[!0-9.]* ]]
 					then
 						timestamp=${command[0]} reflector=${command[1]} seq=${command[3]} rtt_ms=${command[6]} reflector_response=1
 					fi


lynxthecat:
I understand the issue and the fix but the fix is computationally costly because this check has to be applied with every single ping result. 

Isn't there a computationally less expensive check?

Whatever fix is applied I'd like the cost to be kept at an absolute minimum. Because this will incur a performance penalty.

ooonea:
You're right, and I measured it: `[[ =~ ]]` in bash recompiles the ERE on every call, so the regex gate costs ~32 µs per sample on my hardware — versus ~0.3 µs for a bash pattern match doing the same job. Wrong tool on a hot path; amended.

The guard is now a glob character-class check, no regex engine involved:

```bash
# fping
if ((${#command[@]} == 12)) && [[ ${command[6]} != *[!0-9.]* ]]
# ping
if ((${#command[@]} == 9)) && [[ ${command[7]} == time=* && ${command[7]#time=} != *[!0-9.]* ]]
```

Measured over 1M iterations (ratios are what matter, absolute numbers are from an i7): negative-sign check 0.13 µs, this class check 0.49 µs, the old regex 31.8 µs. Against the ~54 µs the read/tokenize/arithmetic path already spends per sample, the class check is under 1% — noise — where the regex would have nearly doubled it.

Why the class check rather than minus-sign only: the crash is any token that can't survive `10#${rtt_us//.}` — the leading minus is just the instance I caught in the wild. A `NaN` in that field aborts the same arithmetic the same way (`printf %.3f` renders it `nan`, then `10#nan` → "value too great for base"), and any stray text that slips through the field count would silently feed garbage into the OWD baselines instead. The class check rejects that whole family at the same order of cost as the sign check. Coverage versus the old regex is identical for anything fping/ping can actually emit — the only extra strings the regex rejected are multi-dot forms like `1.2.3`, which the pingers never print and which don't abort the arithmetic anyway.

Pushed as 291c467; gate matrix re-run: normal fping/ping lines accepted, negative-RTT / unreachable / timed-out lines rejected. `bash -n` and shellcheck unchanged from master.

lynxthecat:
Seem ok to you @rany2?

rany2:
@lynxthecat I think that they use a monotonic clock, I am not convinced that this needed.

lynxthecat:
How so? My understanding from the opening post is that if the system clock is changed fping/ping can report negative values that then crashes out given:

```
$ rtt_ms="-444.1"; printf -v rtt_us %.3f "${rtt_ms}"; (( dl_owd_us=10#${rtt_us//.}/2 ))
bash: ((: dl_owd_us=10#: invalid integer constant (error token is "10#")
```

There might be an easier fix though?

rany2:
@lynxthecat Yes but the post is also AI slop, it might be a hallucinated issue.

rany2:
I cannot reproduce this....

lynxthecat:
It seems this issue might be real:

https://github.com/schweikert/fping/issues/116

https://github.com/schweikert/fping/blob/6731abf450322be7fc1d246fd9f430ec29eba86a/src/fping.c#L124

I don't understand under what exact circumstances the monotonic clock is NOT used.

rany2:
@lynxthecat No it is not, on Linux monotonic clock is used. I also tested iputils-ping. https://github.com/schweikert/fping/blob/6731abf450322be7fc1d246fd9f430ec29eba86a/configure.ac#L16

lynxthecat:
Thanks @rany2. I wonder about virtualization? 

@ooonea what's your actual use case that led to this issue?

rany2:
@lynxthecat Let's just close the PR and ban that user. We should not tolerate this sort of behavior. All they're doing is asking Claude or some other LLM to "Please find issues in this repository and make a PR that pretends you actually came across that issue yourself."

I don't think any kind of virtualization could cause this. The issue you shared predates fping switching to a monotonic clock. That switch happened in early January 2020.

lynxthecat:
@ooonea as you can tell the AI-assistance ruffles feathers. I'm torn because several of your changes have made sense to me, but I'm not as experienced a coder as @rany2 and the time I have to spend on this project is far less than it used to be.

rany2:
FYI https://github.com/search?q=repo%3Aiputils%2Fiputils+REALTIME&type=code. It only sees some usage as a fallback method for random number generation and for the clockdiff tool. Not for iputils-ping.

Please stop making issues up, @ooonea. It wastes our time.

ooonea:
@lynxthecat the use case is a NixOS router on a Starlink line: cake-autorate as a system service, chrony with NTS for time. It happened on its own, in production. From the journal, same second:

```
17:03:58 chronyd: System clock was stepped by -1.759986 seconds
17:03:58 cake-autorate: ERROR: ... ((: dl_owd_us=10#: invalid integer constant (error token is "10#")
17:04:01 systemd: cake-autorate.service: Deactivated successfully.
```

A bare `10#` in that message means the value's `printf %.3f` output began with `-`.

@rany2 you're right about iputils, and I think that is exactly why we differ. iputils **clamps** a backward step: `ping_common.c` warns *"time of day goes back (%ld s), taking countermeasures"* and sets `triptime = 0`. fping does not — `sprint_tm()` has a branch labelled `/* negative (unexpected) */` that formats such a value with `%.2g`, so it prints something like `-1.8e+03`, and the fping arm feeds that straight into `10#`.

On the clock: in my build fping is not on `CLOCK_MONOTONIC`. In `fping.c`, `#if HAVE_SO_TIMESTAMPNS || ONLY_CLOCK_REALTIME` defines `CLOCKID CLOCK_REALTIME` first, and the monotonic block below it is guarded by `#if !defined(CLOCKID)`. `HAVE_SO_TIMESTAMPNS` comes from `AC_CHECK_DECL([SO_TIMESTAMPNS])`, so on Linux the first branch wins unless it is built `--disable-timestamp`. It can be checked without touching any clock: `strace -e trace=setsockopt fping ...` shows `SO_TIMESTAMPNS` being set on the sockets, and that call only compiles under the same macro.

On reproducing it: six backward steps on this box since 1 July, two crashes. It is a race between the step and a sample already in flight, so a clean run is the expected outcome rather than a contradiction.

The smallest useful change is rejecting a leading `-` on the fping arm. The class check I pushed also covers `nan`, which aborts the same arithmetic with a different message, but if you prefer the narrow version I will cut it to that — and if you would rather close this, that is genuinely fine.

Either way, thank you both for the hospitality. I will stop developing against main and keep this work on my own fork; if anyone ever wants something from it, ping me.

rany2:
@ooonea You previously said it was a Debian/Starlink router; and now it's NixOS....

ooonea:
> @ooonea You previously said it was a Debian/Starlink router; and now it's NixOS....

In fact, it used to be Debian, and now it's NixOS. Starlink remains upstream and Protectli downstream; only the operating system has changed.


Schermata del 2026-08-05 17-18-37

rany2:
Can you author the PR descriptions yourself? Just seeing AI slop descriptions doesn't help your case. Also update this PR so we don't have this apply to iputils-ping as it doesn't impact it. Only fping needs this then.

rany2:
BTW can you please reply to comments from others yourself instead of using AI to respond? It's very tiring to respond to a bot.

lynxthecat:
Did that force-push make it in?

https://github.com/lynxthecat/cake-autorate/compare/291c46781b384b7966ee9c6a3032117bd96de2ce..7e73ab7f0d4ac17feea06c54187e72736cfedae5

rany2:
Yes it did, sorry I didn't understand your comment...

lynxthecat:
Thanks for your patience. 

@ooonea I believe your heart is in the right place. So far I am thinking you are a human-AI hybrid rather than pure AI. I'd definitely rather be interacting with those I could share a beer with.

rany2:
@lynxthecat I think so too, I feel like they're just a human that copy-paste AI responses. Personally I find it a bit frustrating and would rather talk to a human, especially because I feel like the AI responses are too verbose and polite (the rude kind of polite). It just doesn't feel genuine.

ooonea:
My life is filled with a thousand things, and for that I am grateful. I use AI because it does many things very well, even though it can make mistakes, sometimes more than humans, sometimes fewer. But ultimately, it allows me to contribute to projects like this one, which I use directly in my everyday life. It speeds up analyses, processes, testing, and much more. Even writing certain responses takes time.
@rany2 thanks for merging it, and for pushing me to keep it only on fping.
@lynxthecat I'm in for the beer. Let me know when you want to get in touch via email.

lynxthecat:
@rany2, years ago I never thought it would be possible to beat the Turing test, but messages like yours above @ooonea makes me wonder. I am genuinely not sure whether it's human or AI. 'My life is filled with a thousand things' is poetic but also feels like an unnatural thing to write, but it could be real. 

To be fair to the Turing test, it would be better to have a real time prompt exchange. I like to think I would always get it right with enough time and enough prompts. 

I asked Grok what to make of the logos without a soul and it came back with the notion that LLMs are no logos as such, and rather a reflection of the logos, which I found at the time quite appealing. Intuitively something does not entirely feel right about it all and I wonder where it will all end up. Presumably introducing yet another wedge between human-human interaction of the sort that might be shared over a beer. 

I am not a fan of technological progress per se - as an old friend said once, 'we are simply wallowing in the excrement of our technological advances'.

ooonea:
Honestly I am a human. The expression comes out of the fact that I am Italian. I don't want to insist trying to convince you that I am human, but I am.

lynxthecat:
I believe you @ooonea. I think you are someone with your heart in the right place using some pretty sophisticated AI tools.

It's just that thinking about the human versus AI dynamic raises some curious philosophical thoughts for me - hence my previous message.

PR #395, opened 2026-08-04
Use irtt streaming mode (-s) for unbounded pinger sessions

### LINKED ISSUE #394: Leverage new irtt streaming mode (-s).
A new streaming mode has been introduced in irtt - see: 

https://github.com/heistp/irtt/issues/45#issuecomment-5150387413

@ooonea I wonder if you might have a shot at a PR that leverages this to replace the existing hack to make irtt work?

Here is the existing hack to make irtt work:

https://github.com/lynxthecat/cake-autorate/blob/73f171d54885b533d5cf2c06aab9677ee58d0088/cake-autorate.sh#L514

Hopefully now we can just pass the continuous output stream on directly:

https://github.com/lynxthecat/cake-autorate/blob/73f171d54885b533d5cf2c06aab9677ee58d0088/cake-autorate.sh#L584

And extract the data elements on the fly:

https://github.com/lynxthecat/cake-autorate/blob/73f171d54885b533d5cf2c06aab9677ee58d0088/cake-autorate.sh#L1457

I'd be grateful if you could follow as closely as possible the existing bash wrappers for ping, fping and tsping.

ooonea:
Closes #394 — and, per the review discussion, gawk is now gone entirely: one `irtt client -s -r --loose -i <interval>` per reflector, parsed in the main loop like every other pinger method.

The raw stream carries neither the reflector nor a wall-clock timestamp, all pingers share the one response fd, and `irtt client` takes exactly one host per session (irtt_client.go:284) with no flag to tag its output — so a per-reflector plain-bash helper prefixes those two fields onto each line and handles the respawn (a server-side max duration clamps the negotiated session and the client just exits; rate-limited to one relaunch per second on quick deaths).

Per-line gating, now in the extraction arm: numeric `rd`/`sd` via glob class checks (no `[[ =~ ]]` on the hot path). That one check drops NaN OWDs (no server timestamps), duplicates (dup=1 lines carry NaN OWDs by construction) and negative OWDs (client and server clocks out of step — which `to_us()` used to reject, and the interim transcoder let through). `ipdv` is legitimately NaN on the first reply of every session and on seqno wrap (heistp confirmed the wrap behavior in heistp/irtt#45), so it never gates a sample. ms→µs uses the same `printf %.3f` idiom as the fping arm.

`--loose`: without it, any server-imposed restriction (e.g. a public server's min send interval) kills the client pre-test at every respawn — `[ServerRestriction] server increased interval from 100ms to 1s`, exit 1, zero samples, silent spin. Reproduced both ways against `irtt server -i 1s`.

Removed: the gawk + time-extension dependencies, the `-d` session bound, the `WaitForPackets` match, the unit-suffix parsing, and `stdbuf` (raw lines come from Go's unbuffered stdout — the old `stdbuf -oL` never had an effect for the same reason). In raw mode irtt routes events/notices to stderr, so stdout carries only sample lines. `ping_prefix_string` now wraps the irtt client itself, as in the other arms (it used to wrap gawk; the popen'd irtt never carried the prefix).

Validation: the gawk checks are gone; a time-bounded probe of `irtt client -h` checks for streaming-mode support — `-s` only exists since irtt 0.9.2, and without the probe an older irtt would reject `-s` at every respawn while the daemon silently produces no samples (min-rate pinning). No pipe to grep (the help goes to stderr and exits 2, which pipefail would turn into a false failure); `read -t` bounds it so a never-exiting irtt can't hang startup.

`irtt_session_duration_m`: deprecated-and-ignored rather than dropped, since dropping the key would break existing config files that set it — happy to drop outright instead, your call from the issue.

Also worth noting from heistp/irtt#45: a 1-hour 1 ms-interval stream (3.6M round trips) shows no leak, RSS stable ~12.5 MB.

Tested on a loopback rig (real irtt 0.9.2 server, wrapper and parser blocks extracted verbatim from the file): 9-field lines end-to-end, µs values matching irtt's own rd/sd within rounding, first sample kept, dup/NaN/negative/short/long lines dropped, respawn paced ~1/s across server restarts, group kill leaves no orphans, stdout stays pure across a server bounce, capability gate passes on 0.9.2 / fails on older / fails-not-hangs on a stuck binary; shellcheck finding count identical to master. Hot-path cost measured over 1M iterations: extraction arm 5.8→20.8 µs/sample (the ms→µs conversion moved in — same class of work the fping arm already does), helper 9.6 µs/line in its own background process vs 4.2 for the gawk stage; at `-i 0.3s` × 6 pingers that is under 0.1% of a core either way. Net diff vs master: −5 lines.

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -514,60 +514,40 @@ start_pinger()
 		irtt)
 			sleep_until_next_pinger_time_slot "${pinger}"
 
+			# The raw stream (-r) carries neither the reflector nor a wall-clock
+			# timestamp, and every pinger shares the one response fd, so a
+			# plain-bash helper prefixes both onto each line; the parsing itself
+			# lives with the other pinger methods in the main loop. A server-side
+			# max duration clamps the negotiated session and the client just
+			# exits, so the helper also respawns the client, rate-limited when
+			# the stream dies quickly (unreachable reflector, refused dial).
+			# --loose accepts server-restricted test parameters (e.g. a public
+			# server's min send interval): without it the client exits pre-test
+			# on any restriction and the respawn loop would spin forever
+			# producing zero samples.
+			local irtt_target=${reflectors[$pinger]}
+			[[ ${irtt_target} == *:* ]] && irtt_target="[${irtt_target}]"
+
 			# shellcheck disable=SC2155  # the subshell status (echo $!) is intentionally unused
 			local pinger_pid=$( (
 				set -m
-				${ping_prefix_string} gawk -v extra_args="${ping_extra_args}" -v interval_s="${reflector_ping_interval_s}" -v duration_m="${irtt_session_duration_m}" -v reflector="${reflectors[$pinger]}" -f <(cat <<-'EOT'
-				@load "time"
-
-				function to_us(val, mult)
 				{
-					# Test bare /s$/ LAST: "5µs" and "5ns" also end in "s", so an
-					# earlier /s$/ branch would swallow them (mult=1000000) and the
-					# µs/ns branches would be unreachable -> sub-ms OWDs inflated 10^6x.
-					mult = (val ~ /ms$/) ? 1000    : \
-						   (val ~ /µs$/) ? 1       : \
-						   (val ~ /ns$/) ? 0.001   : \
-						   (val ~ /s$/)  ? 1000000 : -1
+					while :
+					do
+						session_start_s=${SECONDS}
 
-					if (val ~ /^-/ || mult < 0) return -1
-					return sprintf("%.0f", val * mult)
-				}
+						${ping_prefix_string} irtt client -s -r --loose ${ping_extra_args} -i "${reflector_ping_interval_s}s" "${irtt_target}" |
+							while read -r irtt_line
+							do
+								printf '%s %s %s\n' "${EPOCHREALTIME/.}" "${reflectors[$pinger]}" "${irtt_line}"
+							done
 
-				BEGIN {
-					FS = "[= ]+"
-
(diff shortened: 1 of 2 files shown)


lynxthecat:
But since irtt now outputs a stream like ping, tsping and fping, can't we forego gawk altogether and do what we do with ping, tsping and fping?

Why can't we do something much simpler like:

```
		fping)
			${ping_prefix_string} fping ${ping_extra_args} --timestamp --loop --period "${reflector_ping_interval_ms}" --interval "${ping_response_interval_ms}" --timeout 10000 "${reflectors[@]:0:${no_pingers}}" 2> /dev/null >&"${main_fd}" &
			pinger_pids[0]=${!}
			proc_pids['fping_pinger']=${pinger_pids[0]}
```

I'd like to dispense with gawk assuming it is no longer necessary and bring irtt in line with the ping, fping and tsping implementations.

ooonea:
Done — gawk is gone entirely; the extraction arm now parses the raw stream like ping/tsping/fping (glob class checks, ms→µs via the same `printf %.3f` idiom as the fping arm). PR body updated to match.

One wrinkle kept me from dropping the helper process too: the raw lines carry neither the reflector nor a timestamp, all pingers share the one response fd, and `irtt client` takes exactly one host per session (irtt_client.go:284) with no flag to tag its output — so a per-reflector plain-bash loop still prefixes those two fields and handles the respawn. Plain bash, no gawk, no time extension.

Testing against a real 0.9.2 server surfaced two fixes now included: `--loose` (without it a server restriction, e.g. a public server's min send interval, kills the client pre-test at every respawn — silent zero-sample spin), and the glob gate rejecting negative `rd`/`sd` (unsynced clocks), which `to_us()` used to drop but the interim transcoder let through.

Hot-path numbers are in the PR body; short version: under 0.1% of a core either way at `-i 0.3s` × 6 pingers, and the diff vs master is now −5 lines with no non-ping dependency left on the irtt path.

PR #397, opened 2026-08-13
README: fix the broken star history chart

Dessalines39394:
The star history chart in the Stargazers section is currently broken: the service it relied on can no longer fetch stargazer data, so the chart fails to render. This updates the README to use a working replacement that serves the chart from the same domain, restoring the section without the need for an API token. Please consider merging so the project keeps its Stargazers chart in good shape.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,4 +211,4 @@ sensors are created but never receive data.
 
 ## :stars: Stargazers <a name="stargazers"></a>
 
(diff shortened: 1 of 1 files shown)


lynxthecat:
Thanks - even if you are a bot.

Dessalines39394:
You are welcome, thank you for fixing the URL

lynxthecat:
Out of curiosity, are you a real person? What's your motivation for fixing all these URLs in GitHub repositories?

Dessalines39394:
I am, though I relied on scripts and regexes to get the list of repositories and apply necessary changes. Someone had to do it so I did it

PR #399, opened 2026-08-14
Keep shaper state consistent with tc success

ooonea:
## Summary

- commit an internal shaper rate only after the corresponding `tc qdisc change` succeeds
- derive packet-delay compensation from rates known to be installed in both directions
- propagate `tc` failures, leave failed changes pending for retry, and fail initial setup before starting the monitor
- add focused fake-`tc` regressions for success, failure, retry, mixed-direction failure, and startup failure

## Root cause

`set_shaper_rate()` updated `last_shaper_rate_kbps` and all derived compensation values regardless of the exit status from `tc`. A failed qdisc change therefore made the controller believe the requested rate was active even though the kernel still used the previous rate.

The mismatch also crossed directions: after a failed download update, a successful upload update could recompute both compensations from the pending download value. During startup, failed qdisc configuration could similarly let monitoring begin without a coherent initialized state.

## Impact

The controller now keeps its state aligned with the qdisc, retries failed changes naturally, and exits safely if the initial shaper setup cannot be installed. The rate-selection algorithm is unchanged.

## Tests

- `bash -n cake-autorate.sh tests/test_set_shaper_rate.sh`
- `tests/test_set_shaper_rate.sh`
- ShellCheck on the new test; no new diagnostics versus upstream for the main script
- `git diff --check`

Prepared with assistance from OpenAI Codex.

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -750,38 +750,64 @@ set_shaper_rate()
 {
 	# Fire up tc and update max_wire_packet_compensation if there are rates to change for the given direction
 
-	local direction=${1} # 'dl' or 'ul'
+	local direction=${1} tc_status=0 # 'dl' or 'ul'
 
-	(( shaper_rate_kbps[${direction}] != last_shaper_rate_kbps[${direction}] )) || return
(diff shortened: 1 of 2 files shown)


lynxthecat:
Sorry but I find this too messy.

PR #400, opened 2026-08-14
Preserve fatal exit status for service managers

ooonea:
## Summary

- propagate fatal daemon and worker failures through `launcher.sh` instead of reporting every shutdown as success
- run each cake-autorate instance in a dedicated process group so cleanup also reaches descendants after a main process is killed
- keep operator `SIGINT`/`SIGTERM` shutdowns successful while bounding cleanup of stopped or signal-resistant groups
- require Bash 5.1 for `wait -n -f -p` and add focused process-lifecycle regressions

## Root cause

The launcher waited for all wrappers and then exited successfully, regardless of why an instance stopped. Fatal paths in `cake-autorate.sh` also signalled the main shell without preserving a failure status. As a result, service managers could see a clean exit after an internal fatal error.

Tracking only the immediate wrapper PIDs was insufficient for abnormal termination: if a main process was killed (for example by OOM/SIGKILL), its monitor or pinger descendants could survive. A stopped sibling could also make an unbounded PID-based cleanup wait forever.

## Implementation

- `cake-autorate.sh` records fatal status before requesting shutdown and preserves it through cleanup
- `launcher.sh` starts direct children as dedicated process groups and uses `wait -n -f -p` to identify the first completed instance
- group cleanup is bounded: `TERM`, `CONT`, a grace interval, then `KILL` and a final bounded poll
- reaped children are removed from the active set before other groups are signalled
- normal operator signals remain status 0; fatal child status is returned to the service manager

## Tests

- Bash syntax and ShellCheck (`--severity=warning`)
- full lifecycle suite on Bash 5.3 and Bash 5.1.16
- 200 simultaneous-exit stress iterations
- real-daemon fatal stderr propagation
- killed main process with descendant cleanup
- stopped groups and TERM-resistant worker escalation
- first nonzero, all-zero, and concurrent child exits
- foreground PTY `SIGINT` and normal `SIGTERM`
- no residual test processes; `git diff --check`

The tests run on Linux with GNU Bash. Router-specific runtime validation remains useful during review.

Prepared with assistance from OpenAI Codex.

### DIFF
--- a/INSTALLATION.md
+++ b/INSTALLATION.md
@@ -24,7 +24,8 @@ required tools. To use it:
 
 - [SSH into the router](https://openwrt.org/docs/guide-quick-start/sshadministration)
 
-- Ensure `bash` and `fping` are installed.
+- Ensure Bash 5.1 or newer and `fping` are installed. The launcher relies on
+  Bash's `wait -n -p` support and the installer rejects older Bash versions.
 
   On most OpenWrt installations, you can install them by running:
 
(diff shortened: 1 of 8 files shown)


lynxthecat:
Again I'm sorry but this is far too convoluted and messy.

PR #414, opened 2026-08-25
Experiment: reduce buffered log collector CPU using per-record framing

lynxthecat:
## Outcome

> [!NOTE]
> This is an archived experiment and is not recommended for merging.

The implementation correctly reduced CPU use for saturated log streams, but later paced testing showed that per-record framing increased CPU use at the rates representative of cake-autorate. It is retained here as a record of the investigation. PR #415 explored fixed padding as a follow-on alternative and reached the same overall conclusion.

## Motivation

The buffered log-file collector consumed log data using:

```bash
read -N "${log_file_buffer_size_B}" \
	-t "${log_file_buffer_timeout_s}"
```

Combining an exact-length read with a timeout makes Bash's data-consuming path expensive under continuously available input. This experiment replaced that timed read with a small internal framing protocol so that the collector could consume headers and payloads using untimed exact reads.

## Experimental implementation

Each accepted formatted log record is written atomically to the internal pipe as:

```text
Ldddd<payload>
```

`Ldddd` is a fixed five-byte header whose four decimal digits declare the payload length.

The implementation:

- accepts payloads up to 4091 bytes so that each complete frame remains within Linux's 4096-byte `PIPE_BUF`;
- reads and validates an untimed five-byte header, followed by an untimed exact payload read;
- accumulates payloads in the existing file-output buffer;
- uses a periodic `L0000` control frame from `log_file_waker` at the configured 500 ms default interval;
- writes a non-empty partial buffer on `L0000`, while still performing rotation and signal checks when the buffer is empty;
- excludes framing headers and control frames from the log file; and
- rejects oversized records before any partial pipe write.

Final head `c874d15` skips an empty file-output operation on padding-only wakes. It does not alter the ordinary per-record framing work.

## CPU benchmarks

The initial unpaced burst benchmarks showed substantial gains because the original timed read continuously consumed available bytes. A subsequent controlled benchmark used one unreported warm-up and five measured runs, alternated old/new ordering, and verified exact output byte counts.

Paced results used approximately 109-byte records and report median combined producer, collector and waker CPU per 60 seconds:

| Workload | Original timed collector | Framed collector | Difference |
|---|---:|---:|---:|
| Idle | 0.02 s | 0.03 s | +0.01 s |
| 2 records/s | 0.05 s | 0.08 s | +0.03 s |
| 5 records/s | 0.10 s | 0.14 s | +0.04 s |
| 10 records/s | 0.18 s | 0.26 s | +0.08 s |
| 20 records/s | 0.33 s | 0.47 s | +0.14 s |
| 50 records/s | 0.77 s | 1.05 s | +0.28 s |
| 100 records/s | 1.29 s | 1.83 s | +0.54 s |

No steady-rate break-even was observed at or below 100 records/s.

Unpaced burst results were:

| Workload | Original | Framed | Difference |
|---|---:|---:|---:|
| 30,000 × 109-byte records | 1.82 s | 1.56 s | 14.3% lower |
| 4,000 × 4091-byte records | 7.49 s | 0.98 s | 86.9% lower |

The controlled paced benchmark was performed against preceding framed revision `f781539`. The final `c874d15` change only removes an empty output operation on `L0000`; it does not remove the header formatting, header read, validation, decimal parsing or separate payload read responsible for the paced regression.

## Interpretation

Framing removes timeout processing from reads that consume log bytes, but adds work for every ordinary record. That trade-off is favourable under saturation and unfavourable for evenly paced short records. Normal cake-autorate logging is paced, with approximately 20 accepted reflector responses/s under the default six-pinger, 0.3-second configuration.

The experimental mechanism is therefore not suitable as a general CPU optimisation.

## Validation

Validation included:

- `bash -n cake-autorate.sh`;
- `git diff --check`;
- ShellCheck 0.11.0 at warning severity, normally and with sourced files followed;
- concurrent atomic frame writers;
- oversized-record rejection;
- low-volume wake and flush behaviour;
- signal handling;
- rotation, export, reset and graceful-shutdown smoke tests; and
- verification that an empty `L0000` wake does not perform an empty file write.

GitHub reports no repository CI checks for this PR.

------

[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a8de104ed6c832e8ab808af6cddd2e4)

### DIFF
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -390,7 +390,6 @@ flush_log_pipe()
 
 log_file_waker()
 {
-	printf -v log_file_buffer_timeout_s %.3f "${log_file_buffer_timeout_ms}e-3"
 	while kill -0 "${proc_pids['maintain_log_file']}" 2>/dev/null
 	do
 		sleep_s "${log_file_buffer_timeout_s}"
@@ -1238,6 +1237,7 @@ then
 		log_file_max_time_s=log_file_max_time_mins*60,
 		log_file_max_size_bytes=log_file_max_size_KB*1024
 	))
+	printf -v log_file_buffer_timeout_s %.3f "${log_file_buffer_timeout_ms}e-3"
 	exec {log_fd}<> <(:)
 	maintain_log_file &
 	proc_pids['maintain_log_file']=${!}
--- a/cake-autorate.sh
+++ b/cake-autorate.sh
@@ -115,6 +115,7 @@ cleanup_and_killall()
 
 	((terminate_maintain_log_file_timeout_ms=log_file_buffer_timeout_ms+500))
 	terminate "${proc_pids['maintain_log_file']:-}" "${terminate_maintain_log_file_timeout_ms}"
+	terminate "${proc_pids['log_file_waker']:-}"
 
 	[[ -d ${run_path} ]] && rm -r "${run_path}"
 	rmdir /var/run/cake-autorate 2>/dev/null
@@ -154,7 +155,7 @@ log_msg()
 {
 	# send logging message to stdout, log file fifo, log file and/or system logger
 
-	local type=${1} msg=${2} instance_id=${instance_id:-"unknown"} log_timestamp=${EPOCHREALTIME}
+	local type=${1} msg=${2} instance_id=${instance_id:-"unknown"} log_timestamp=${EPOCHREALTIME} log_frame_payload
 
 	case ${type} in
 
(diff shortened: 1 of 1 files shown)


lynxthecat:
Closing without merge.

The original saturated-stream benchmarks correctly identified CPU cost associated with Bash's timed data-consuming `read`, but subsequent paced benchmarks showed that per-record framing does not improve the workload relevant to normal cake-autorate operation.

In the later controlled benchmark, framing used more combined producer, collector and waker CPU at every evenly paced rate tested from 2 to 100 approximately 109-byte records/s. At 20 records/s, median CPU increased from 0.33 to 0.47 seconds per 60 seconds; at 100 records/s it increased from 1.29 to 1.83 seconds.

Framing remained beneficial under unpaced saturation: it used 14.3% less median CPU for 30,000 approximately 109-byte records and 86.9% less for artificial near-maximum records. Normal cake-autorate output is, however, paced rather than continuously saturated. At those rates, formatting, reading, validating and parsing a header for every record, followed by a separate payload read, outweighed the timed-read saving.

The paced benchmark was performed against a preceding framed revision. The final `c874d15` commit only avoids an empty file-output operation on an `L0000` wake and does not remove the ordinary per-record framing costs responsible for the regression.

PR #415 investigated fixed padding as a way of retaining untimed reads without per-record headers. OpenWrt testing likewise found that its fixed costs outweighed its benefit for sparse and SUMMARY-only logging.

The framing implementation and its investigation are retained here for reference, but the original timed collector remains preferable overall.

